PR #25, opened 2022-05-05
Add German translation

Abestanis:
This adds German translations.

The pull request depends on #24 and should be merged after it, therefore I'm keeping it as a draft. Unfortunately, I cannot set the base branch to the branch of #24, so this pull request contains all commits from that pull request as well. They will disappear, once #24 is merged.

This pull request is ready to be reviewed, the only change is commit eea402768c04043da9ccd1df553a275171b2e81f,


Abestanis:
@nt4f04uNd I wonder if it would be ok with you to remove the `"@..."` metadata entries for all translations that are a text type and have no parameters or other metadata. The text type seems to be the default, so these metadata entries are not actually required.

They bloat up the translation files by a lot, making them bigger and harder to navigate.

If you are ok with it, I would remove them from the German translation in this pull request and create a new pull request to remove them in the English and Russian translation files as well.

And while we are on the topic, how do you feel about removing all unused translations? We could always retrieve them later from the git history, if we need them.

nt4f04uNd:
It seems fine to remove the unused metadata, but let's first merge https://github.com/nt4f04uNd/sweyer/pull/24, so those changes don't add more mess to the review.

Let's keep the unused strings, I kept them to reuse in the future. One day I will review them and delete, when I'm sure they are no longer needed.

PR #28, opened 2022-05-05
Add tests for the app exit confirmation behavior

Abestanis:
This adds test for the app exit confirmation behavior and the setting introduced in #26. Tests for the new settings screen are not included.

I'm not super familiar with Flutter testing, if you have an ideas on how to improve these, I would be happy to hear them.
I'm also not supper happy about hard-coding the flutter toast package method channel name and the method names, but I don't see any other way.

I also manually tested that switching from `DateTime` to the to the `clock` package doesn't change how the back button confirm timeout works.


test/observer/app.dart
```diff
@@ -0,0 +1,30 @@
...
+      }
+      return null;  // Ignore unimplemented method calls
+    });
+  }
+
+  /// Execute the provided [callable] and return whether
+  /// an application close request was made by it. 
+  Future<bool> haveCloseRequest(Future<void> Function() callable) async {
+    _haveCloseRequest = false;
+    await callable();
```
nt4f04uNd, line 27: This does not need to have a callable, this class can just expose the number of times SystemNavigator.pop was called, and then we can check that number in tests that need this

test/observer/app.dart
```diff
@@ -0,0 +1,30 @@
+import 'package:flutter/services.dart';
+
+import '../test.dart';
+
+/// An observer for the application state. 
+class AppObserver {
```
nt4f04uNd, line 6: Rename this (and files, and the other observer) to *ChannelObserver

This one would be SystemChannelsObserver

test/observer/toast.dart
```diff
@@ -0,0 +1,47 @@
...
+    });
+  }
+
+  /// Verify that the provided [callable] shows a toast.
+  /// If provided, also verify that the toast showed the given [message].
+  /// If no toast was shown or the message doesn't match,
+  /// fail with the given [reason].
+  Future<void> expectShowsToast(Future<void> Function() callable, {String? message, String? reason}) async {
+    lastToastMessage = null;
+    await callable();
```
nt4f04uNd, line 30: It should rather record a log of shown toasts, which it then exposes and we check it in our tests

Therefore I believe that we don't need `expectShowsToast` and `expectShowsNoToast` at all

test/routes/home_route_test.dart
```diff
@@ -69,4 +72,75 @@ void main() {
       expect(ContentControl.instance.state.allSongs.songs, [songWith()]);
     });
   });
+
+  testWidgets('app shows exit confirmation toast', (WidgetTester tester) async {
```
nt4f04uNd, line 76: Can we rather have:
* one test that verifies that for `true` - both that the app shows toast and doesn't call pop with setting set to true, and only allows to pop after a certain duration
* and the other test that checks that with `false` we immediately are popped

I like such structure more, because it tests the entire feature flow, as if we are looking at it from a user's perspective, and it could be more clearely formalized, if we were writing a test checklist, for example

test/routes/home_route_test.dart
```diff
@@ -69,4 +72,75 @@ void main() {
...
+          () => toastObserver.expectShowsToast(
+            BackButtonInterceptor.popRoute,
+            message: l10n.pressOnceAgainToExit,
+            reason: 'Expected the app to ask for confirmation before exiting',
+          ),
+        ),
+        isFalse,
+        reason: 'Expected the app not to close after showing the toast',
+      );
+      await tester.binding.delayed(const Duration(seconds: 5));
```
nt4f04uNd, line 92: Move the duration to constant in config.dart and test that exactly after that duration it allows us to quit the app

nt4f04uNd (commented):
> I also manually tested that switching from DateTime to the to the clock package doesn't change how the back button confirm timeout works.

It's how it should be done, correct

test/observer/app.dart
Abestanis: The idea was to ensure that the request happened in the given callable and not before. I thought it would look better than having

```dart
appObserver.clearCloseRequests();
// Do something
expect(appObserver.hadCloseRequest(), true);
```

What do you think?

Abestanis: Same reasoning for `expectShowsToast` and `expectShowsNoToast`.

nt4f04uNd: I can see the idea behind this, but I have never seen tests like this. It's more intuitive that we write

```dart
expect(systemChannelsObserver.popCount, 0);
callSomethingThatTriggersPop();
expect(systemChannelsObserver.popCount, 1);
```

test/routes/home_route_test.dart
```diff
@@ -69,4 +72,75 @@ void main() {
...
+      expect(
+        await appObserver.haveCloseRequest(
+          () => toastObserver.expectShowsToast(
+            BackButtonInterceptor.popRoute,
+            message: l10n.pressOnceAgainToExit,
+            reason: 'Expected the app to ask for confirmation before exiting after the previous message timed out',
+          ),
+        ),
+        isFalse,
```
nt4f04uNd, line 101: also, for booleans prefer use `false` and `true`, these matchers are essentially the same

test/observer/app.dart
Abestanis: I opted for an approach where the getter clears the property on access. This is produces the cleanest test code, and should be robust enough against reordering of steps in the test.

test/routes/home_route_test.dart
Abestanis: Done

test/routes/home_route_test.dart
Abestanis: Good idea, done.

test/routes/home_route_test.dart
Abestanis: I also tested if using `equals` is better, but the output is almost the same, there is no real value added by it, so I decided to not use it for the toast string comparisons.

test/observer/app.dart
Abestanis: Done, it's more descriptive. 👍

test/observer/toast.dart
```diff
@@ -0,0 +1,32 @@
+import 'package:flutter/services.dart';
+
+import '../test.dart';
+
+/// An observer for toast messages from the flutter toast package.
+class ToastChannelObserver {
+  /// The method channel used by the flutter toast package
+  static const MethodChannel _channel = MethodChannel('PonnamKarthik/fluttertoast');
+  Map<String, dynamic>? _lastToastArguments;  /// The arguments for the last requested toast message.
```
nt4f04uNd, line 9: Why only last toast and not a list of all?

test/observer/toast.dart
```diff
@@ -0,0 +1,32 @@
...
+    clearLastToast();
+    return lastArguments?['msg'] as String?;
+  }
+  
+  /// Create a new toast channel observer, which automatically
+  ///  unregisters any previously created observer.
+  ToastChannelObserver(WidgetTester tester) {
+    tester.binding.defaultBinaryMessenger.setMockMethodCallHandler(_channel, (call) {
+      if (call.method == 'showToast') {
+        _lastToastArguments = Map.castFrom(call.arguments);
```
nt4f04uNd, line 21: nit: seems like it can be `as Map<String, dynamic>?`

test/observer/toast.dart
```diff
@@ -0,0 +1,32 @@
...
+        return null;
+      }
+      return null;  // Ignore unimplemented method calls
+    });
+  }
+
+  /// Forget the last received toast.
+  void clearLastToast() {
+    _lastToastArguments = null;
```
nt4f04uNd, line 30: This can be removed

test/routes/home_route_test.dart
```diff
@@ -67,4 +71,43 @@ void main() {
...
+  testWidgets('app shows exit confirmation toast if enabled in the preferences', (WidgetTester tester) async {
+    await Prefs.confirmExitingWithBackButton.set(true);
+    await tester.runAppTest(() async {
+      final SystemChannelObserver systemObserver = SystemChannelObserver(tester);
+      final ToastChannelObserver toastObserver = ToastChannelObserver(tester);
+      await BackButtonInterceptor.popRoute();
+      expect(toastObserver.lastToastMessage, l10n.pressOnceAgainToExit,
+          reason: 'Expected the app to ask for confirmation before exiting');
+      expect(systemObserver.closeRequests, 0,
+          reason: 'Expected the app not to close after showing the toast');
```
nt4f04uNd, line 84: While it might seem good have a human readable sentence to the failure, to me it clutters the test a bit, since it's faster to just look at the expect to understand what went wrong

Can we get rid of the `reason:` everywhere, what do you think? If you disagree, it's ok if we keep them :-)

test/observer/system.dart
```diff
@@ -0,0 +1,31 @@
...
+      if (call.method == 'SystemNavigator.pop') {
+        _closeRequests++;
+        return null;
+      }
+      return null;  // Ignore unimplemented method calls
+    });
+  }
+
+  /// Reset the number of recorded close requests to zero.
+  void clearCloseRequest() {
```
nt4f04uNd, line 28: This can be removed

test/observer/toast.dart
```diff
@@ -0,0 +1,32 @@
...
+import '../test.dart';
+
+/// An observer for toast messages from the flutter toast package.
+class ToastChannelObserver {
+  /// The method channel used by the flutter toast package
+  static const MethodChannel _channel = MethodChannel('PonnamKarthik/fluttertoast');
+  Map<String, dynamic>? _lastToastArguments;  /// The arguments for the last requested toast message.
+  String? get lastToastMessage {
+    final lastArguments = _lastToastArguments;
+    clearLastToast();
```
nt4f04uNd, line 12: Why do we need to clear it here (same in the other observer)?

test/observer/toast.dart
Abestanis: Being able to clear the last toast message is very important in my opinion. Consider this:

``` dart
... // Code that should generates Toast 1.
expect(toastObserver.lastToastMessage, l10n.toast1);
... // Code that should generate Toast 2.
expect(toastObserver.lastToastMessage, l10n.toast2);
```

If we don't clear the `lastToastMessage` after checking for `toast1`, it leaks in the second part of the test. If the code that should generate `toast2` doesn't generate a toast, we wan't the second expect to fail and tell us that no toast was displayed, which is unequal to the expected `toast2`.
Instead, we get `toast1` is unequal to `toast2`, which is confusing and doesn't point us in the direction of what's going wrong here. Even worse, consider if `toast1` and `toast2` were actually the same string, for example if I want to test that the exit confirmation toast shows again when I wait 2 or more seconds. In that case, even if the second code doesn't generate the toast, the except still passes, because `lastToastMessage` contains the `toast1` from the first code. So the test becomes useless and even harmful at this point, because we think we are testing something that we are actually not.

I strongly believe that we must clear in order to protect us from situations like this. Now, one might argue that we could solve this her by just keeping a list of all recorded toast messages. But this only works for this case because we are handling strings, and most importantly, this makes the test more complicated and dependent on what was done before. If we decides to add an additional part before this test, the history of toasts changes. But in our test, we are only ever interested about the history since we last observed last, or since we last cleared.

test/routes/home_route_test.dart
Abestanis: I will examine where they actually add any value.

test/observer/toast.dart
Abestanis: See my comment above for reasoning.

test/observer/toast.dart
Abestanis: We could do this, seems a bit ugly though. What do you think?
```suggestion
        _lastToastArguments = Map.castFrom<dynamic, dynamic, String, dynamic>(call.arguments);
```

test/observer/toast.dart
nt4f04uNd: I believe the approach you have chosen with keeping only last message is error-prone and less clear.

> But this only works for this case because we are handling strings, and most importantly, this makes the test more complicated and dependent on what was done before. 

This would work with any matcher, not only strings, it doesn't make the test any more complicated. 

> If we decides to add an additional part before this test, the history of toasts changes. But in our test, we are only ever interested about the history since we last observed last, or since we last cleared.

That would be perfect if we notice that our test is affected by something we didn't notice on our own and change the test accordingly to expect it, or investigate why this is and fix what affected it.

test/routes/home_route_test.dart
Abestanis: Done, see cf334940c306eb4eb5e88a90d79efc4fa31be798.

test/observer/toast.dart
nt4f04uNd: Also this kind of magic of clearing the state after calling a getter is what should be avoided in any kind of code

Abestanis: > I believe the approach you have chosen with keeping only last message is error-prone and less clear.

I disagree, the current approach avoids potential errors like the one I showed above. If we keep the history, I need an additional check to detect the fact that there was no toast generated:
```dart
... // Code that should generates Toast 1.
expect(toastObserver.toastMessages.last, l10n.toast1);
var numToastsBefore = toastObserver.toastMessage.length;
... // Code that should generate Toast 2.
expect(toastObserver.toastMessage.length, numToastsBefore + 1);
expect(toastObserver.lastToastMessage, l10n.toast2);
```

This not only makes the test more complicate to read, it also generates an additional burden for us when we write the tests, because we need to remember to add this check, otherwise our tests will not test what we actually think they test.

Abestanis: > That would be perfect if we notice that our test is affected by something we didn't notice on our own and change the test accordingly to expect it, or investigate why this is and fix what affected it.

The problem is that if we do that, we are not testing that the app behaves as expected, we are testing that out test behaves exactly as we originally thought it would. The failing tests don't indicate a problem in the app, just a change in the test. I don't see the benefit in that, because it makes our tests less flexible, and it moves our focus away from the app to the test. It could also cause us to think that a failing test is caused by a change we do to the test, when in reality it uncovers a bug in the actual app, but we then "fix" the test, because we think the failure is caused by our change in the test before.

Abestanis: > Also this kind of magic of clearing the state after calling a getter is what should be avoided in any kind of code

I do agree with that, how about this:

I remove the `lastToastMessage` getter and instead add a function `popLastToastMessage`. Then the intent would be clear.

nt4f04uNd: > The problem is that if we do that, we are not testing that the app behaves as expected, we are testing that out test behaves exactly as we originally thought it would.

No, we are cementing in a test that our app shown exactly those toasts that we expected

```dart
expect(toastObserver.toastMessagesLog, []);

functionThatShowsOneToast();
expect(toastObserver.toastMessagesLog, [l10n.toast1]);

someFunctionThatShouldOnlyShowOneMoreAdditionalToast();
expect(toastObserver.toastMessagesLog, [l10n.toast1, l10n.toast2]);
```

If at some point `someFunctionThatShouldOnlyShowOneMoreAdditionalToast` starts to show N additional toasts that we didn't expect, with your approach we fail to notice that. Checking it like above will make the test fail and we will investage why is that - if expected, change the test accordingly, otherwise fix `someFunctionThatShouldOnlyShowOneMoreAdditionalToast`.

I insist that it's done like this.

test/observer/toast.dart
nt4f04uNd: Tohse types are inferred by language, it's not needed

test/observer/toast.dart
Abestanis: Ok, I see now that your approach is definitely the only way to deal with that approach, and it does make the tests more explicit, which they should be. I made the changes in the latest commit.

nt4f04uNd:
This is much better now, thank you!

PR #31, opened 2022-05-12
Improve the permission system tests

Abestanis:
This fixes an error, which only happens in the debug version of the app. The following assert is triggered, when successfully granting storage permission:
https://github.com/nt4f04uNd/sweyer/blob/5e18cda1a292cb7deb04c5587604c4b4c32ad331/lib/logic/control.dart#L44
<details>
<summary>Stacktrace</summary>

```java
E/flutter: [ERROR:flutter/lib/ui/ui_dart_state.cc(209)] Unhandled Exception: 'package:sweyer/logic/control.dart': Failed assertion: line 44 pos 12: '_disposed.value': is not true.
    #0      _AssertionError._doThrowNew (dart:core-patch/errors_patch.dart:51:61)
    #1      _AssertionError._throwNew (dart:core-patch/errors_patch.dart:40:5)
    #2      Control.init (package:sweyer/logic/control.dart:44:12)
    #3      ContentControl.init (package:sweyer/logic/player/content.dart:202:11)
    #4      Permissions.requestClick (package:sweyer/logic/permissions.dart:24:37)
    <asynchronous suspension>
    #5      _NoPermissionsScreenState._handlePermissionRequest (package:sweyer/routes/home_route/home_route.dart:194:5)
    <asynchronous suspension>
    #6      NFButton._handleOnPressed (package:nt4f04unds_widgets/src/widgets/buttons.dart:194:17)
    <asynchronous suspension>
```
</details>
<details>
<summary>This causes the app to get stuck.</summary>

Error after permission grant
</details>

The error is caused, because the `ContentControl` is reinitialized after the permission was granted in
https://github.com/nt4f04uNd/sweyer/blob/5e18cda1a292cb7deb04c5587604c4b4c32ad331/lib/logic/permissions.dart#L24
But the assert prevents the reinitialization.

This improves the test so it covers the code of the assertion and shows the problem. In general, this improves the test coverage of the permission screen, by faking/mocking less and testing the failure paths too.

Note: The last commit includes part of #28, because I needed to test the toast behavior.


lib/logic/control.dart
```diff
@@ -41,7 +41,6 @@ abstract class Control {
   /// Must be called when the control needs to be initialized.
   @mustCallSuper
   void init() {
-    assert(_disposed.value);
```
nt4f04uNd, line 44: Just removing this is incorrect, this is asserting that controls are not initialized twice. This is not ideal however, and we should rather silently ignore additional `init` calls in subclasses, by checking the `disposed`.

I'm not sure why you changed the rest of the code in this pull request, can you clarify?

test/observer/permissions.dart
```diff
@@ -0,0 +1,79 @@
...
+import 'package:permission_handler_platform_interface/permission_handler_platform_interface.dart';
+
+import '../test.dart';
+
+/// An observer for platform permission requests.
+class PermissionsChannelObserver {
+  /// The method channel used by the flutter permissions package
+  static const MethodChannel _channel = MethodChannel('flutter.baseflow.com/permissions/methods');
+  final List<Permission> _requestedPermissions = [];  // The permissions which were requested.
+  List<Permission> get requestedPermissions  => UnmodifiableListView(_requestedPermissions);
```
nt4f04uNd, line 14: here and for other getters - move private variable under the getter and make the comment to be a doc comment of the getter

also add a newline between different getters

example:

```dart
  /// The permissions which were requested.
  List<Permission> get requestedPermissions  => UnmodifiableListView(_requestedPermissions);
  final List<Permission> _requestedPermissions = [];

  /// ....
```

test/observer/permissions.dart
```diff
@@ -0,0 +1,79 @@
...
+    } else {
+      setPermissionResolvable(permission, () async => status);
+    }
+  }
+}
+
+extension PermissionObserverExtension on WidgetTester {
+  /// Create a new PermissionsChannelObserver, which overwrites the default one
+  /// for this test.
+  PermissionsChannelObserver overwritePermissionObserver() => PermissionsChannelObserver(binding);
```
nt4f04uNd, line 78: Why do we even need this? Why not just `permissionsObserver = PermissionsChannelObserver()`;

test/observer/permissions.dart
```diff
@@ -0,0 +1,79 @@
...
+  /// Whether an attempt to open the settings should succeed.
+  bool isOpeningSettingsSuccessful = true;
+  
+  /// Permissions whose values are specified, all other permissions are implicitly granted.
+  final Map<Permission, Future<PermissionStatus> Function()> _specifiedPermissions = {};
+
+  /// Create a new permissions observer, which automatically
+  /// unregisters any previously created observer.
+  PermissionsChannelObserver(TestWidgetsFlutterBinding binding) {
+    binding.defaultBinaryMessenger.setMockMethodCallHandler(_channel, (call) async {
```
nt4f04uNd, line 29: Please refactor this (and other places as such) to just 

```dart
const MethodChannel('flutter.baseflow.com/permissions/methods').setMockMethodCallHandler((call) async {
// ...
});
```

test/observer/permissions.dart
Abestanis: Actually, the comment is already for the getter, at least my IDE displays for the getter and not for the variable. But I agree that your suggestion is cleaner.

test/observer/permissions.dart
Abestanis: Unfortunately, extensions can't declare fields.

Abestanis: Ah, if you mean why we need the extension at all, it's to avoid having to write:
```dart
final permissionsObserver = PermissionsChannelObserver(tester.binding);
```

So it's not necessary, but it makes the code look better and also self documents whats happening (overwriting the default permission observer).

test/observer/permissions.dart
Abestanis: No, this way is deprecated and [was removed in Flutter 2.5](https://docs.flutter.dev/release/breaking-changes/mock-platform-channels), the it is currently is the new [recommended way](https://stackoverflow.com/a/68240542).

nt4f04uNd: I'm on Flutter 2.10.3 and is still working for me, in fact, there are places in the app that use this syntax. I would not take an SO answer as a "recommended way". While the current comment on this method says that it should be deprecated, we should not care until it really is deprecated - maybe it will not ever be deprecated.

test/observer/permissions.dart
nt4f04uNd: That's an inherentent action of creating an observer, so I'm for removing this method, since it's an equivalent of a constructor call

test/observer/permissions.dart
Abestanis: It's true, but the only reason why this still works is because of `TestMethodChannelExtension`, which is defined in `fultter_test/lib/src/deprecated.dart` and whose documentation clearly advertises against it's usage.

nt4f04uNd: OK, this makes sense. Let's refactor everything else to use the new binder methods then.

lib/logic/control.dart
Abestanis: @nt4f04uNd Is this ok with you, or do you want another solution?

nt4f04uNd: It is OK

PR #32, opened 2022-05-12
Integrate codecov

Abestanis:
I think it would be nice be able to see code coverage online with [codecov](https://about.codecov.io). The only thing left to do for this to work is to create an account named `nt4f04uNd` with codecov and create a projecte named `sweyer`. Uploading coverage information should just work after that. There is no need to add a token to GitHub for public projects.

When the first coverage is uploaded we should check if we need the [dart-full-coverage action](https://github.com/marketplace/actions/dart-full-coverage). It sounds like by default files not loaded during the test don't count towards the coverage result? We'll have to see if that's true.

### DIFF
--- a/.github/workflows/sweyer.yml
+++ b/.github/workflows/sweyer.yml
@@ -13,8 +13,5 @@ jobs:
         with:
           testing_arguments: --coverage
 
-      - name: 📊 Check Code Coverage
-        uses: VeryGoodOpenSource/very_good_coverage@v1
-        with:
-          path: coverage/lcov.info
-          min_coverage: 0
+      - name: ☂ Upload Code Coverage
+        uses: codecov/codecov-action@v3
--- a/.gitignore
+++ b/.gitignore
@@ -84,8 +84,11 @@ untranslated-messages.txt
 
 .fvm/flutter_sdk
 
-# ignore golden failure outputs
+# Ignore golden failure outputs
 **/failures/*.png
+
+# Coverage results
+coverage
 
 .project
 .classpath
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -53,3 +53,9 @@ golden files:
 
 If you left the auto-commit option unchecked, those new files can be put into `test/golden/goldens`
 folder and manually pushed to your PR.
+
+## Coverage [![codecov](https://codecov.io/gh/nt4f04uNd/sweyer/branch/master/graph/badge.svg)](https://codecov.io/gh/nt4f04uNd/sweyer)
+
+Coverage measures which lines of code are covered during the tests. It shows which part of the
+code is untested. When adding a new feature, a test must be added to verify the new feature and to
+avoid reducing the test coverage.
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,0 +1,1 @@
+comment: false


README.md
```diff
@@ -1,3 +1,7 @@
+[![codecov](https://codecov.io/gh/nt4f04uNd/sweyer/branch/master/graph/badge.svg)](https://codecov.io/gh/nt4f04uNd/sweyer)
+
+# Sweyer
```
nt4f04uNd, line 3: I don't like how it looks - the current README is perfect to me, so let's just not add this badge anywhere

nt4f04uNd:
> The only thing left to do for this to work is to create an account named nt4f04uNd with codecov and create a projecte named sweyer

Seems like it's already working, so I didn't even needed to authorize it to my account

> When the first coverage is uploaded we should check if we need the [dart-full-coverage action](https://github.com/marketplace/actions/dart-full-coverage)

Not sure, it generates an OK coverage, it seems, but it would worth to experiment with it - you can try it here. It would also be cool if the coverage somehow was shown in the list of github checks, so we don't have to go to Codecov to manually check it.

README.md
Abestanis: Hm, I would like to have a link to the coverage statistics, and I think it would also be helpful to someone who wants to participate. How about we move the badge into the `CONTRIBUTING.md` and I add a short paragraph about coverage?

Abestanis:
> Seems like it's already working, so I didn't even need to authorize it to my account

Awesome!

> Seems like it's already working, so I didn't even need to authorize it to my account

It looks like it is required, the `lib/widgets/debug.dart` does not appear.

> It would also be cool if the coverage somehow was shown in the list of github checks [...]

Let me see what I can find. I know that the action can comment coverage results into the pull request, but I always thought that was very noisy and hard to read.

Abestanis:
Adding the `dart-full-coverage` action has resolved the problems of the missing files in the report. 💯

README.md
nt4f04uNd: Yes, `CONTRIBUTING.md` would be a good place for it

Abestanis:
Do we have any randomness in our tests? The coverage of the last two commits have a difference in `lib/logic/player/music_player`, even though the code hasn't changed:

[Commit 1](https://app.codecov.io/gh/nt4f04uNd/sweyer/blob/74e68cf77a4f42d32bd7927ffe79650f0fe50291/lib/logic/player/music_player.dart) vs [Commit 2](https://app.codecov.io/gh/nt4f04uNd/sweyer/blob/d28e95be251bd286c6d636c3f3499426fd5eeaf5/lib/logic/player/music_player.dart)
image

nt4f04uNd:
I would take a wild guess and say that golden tests execute a little different each time, hence some code like this might not get executed sometimes. Just a guess though.

Abestanis:
@nt4f04uNd So it looks like commenting is enabled by default. What do you think of this comment, for me it doesn't really look nice. I would disable it, and hope that it still adds these status checks. It seems to be very slow, but there is the option to add it as an app, maybe that will speed it up: https://github.com/apps/codecov

Abestanis:
It does still create the status checks, even when the comment is disabled:
image

We can even configure them: https://docs.codecov.com/docs/commit-status

This could also replace the `VeryGoodOpenSource/very_good_coverage`, we could set it up that degrading the coverage is not allowed.

nt4f04uNd:
> What do you think of this comment, for me it doesn't really look nice

Doesn't seem to be useful, so I'm for disabling it

> This could also replace the VeryGoodOpenSource/very_good_coverage, we could set it up that degrading the coverage is not allowed.

It's nice to know the coverage but it should not be enforced, I don't want it to affect my workflow.

Seems like we could remove very_good_coverage, since it does the same thing as codecov.

Abestanis:
#35 should be merged into `master` and then into `1.0.8`, I will then rebase this and mark it as ready.

nt4f04uNd:
Merged master into 1.0.8

PR #35, opened 2022-05-15
Add a dedicated Golden update workflow

Abestanis:
This adds a dedicated workflow that can be triggered from the Actions tab on GitHub to update the Golden test artifacts on a branch. This required splitting the existing workflow for reusabillity. The workflow uploads the updated test artifacts, but it can also commit the changes directly to the repository:
Update Goldens Workflow
Update Goldens Workflow 2
image

I had to make this pull request against `master`, because the default branch must have the workflow file, otherwise it won't show up in the Actions tab. To use this workflow, the branch must also have the workflow file, so it would be nice if we could merge `master` into `1.0.8` after this is merged.

To prevent just everyone from being able to trigger golden updates and auto commits, I added a check a commit was made on the branch under the same name as the username of the user running the workflow. This should work, as long as people have the same git and GitHub username. We could also prevent auto commits on master.


.github/workflows/update_goldens.yml
```diff
@@ -0,0 +1,42 @@
+name: Update Goldens
+on: 
+  workflow_dispatch:
+    inputs:
+      commit:
+        type: boolean
+        default: true
+        required: true
+        description: Commit the new Goldens to this branch.
```
nt4f04uNd, line 9: unnecessary "the"

.github/workflows/update_goldens.yml
```diff
@@ -0,0 +1,42 @@
...
+      contents: write
+    steps:
+      - name: 📚 Git Checkout
+        uses: actions/checkout@v3
+
+      - name: 🏗 Build and test Sweyer
+        uses: ./.github/workflows/build_and_test
+        with:
+          flutter_channel: stable
+          flutter_version: 2.10.5
```
nt4f04uNd, line 24: let's default these two parameters into the script

.github/workflows/update_goldens.yml
```diff
@@ -0,0 +1,42 @@
...
+      - name: 📁 Save updated Golden test artifacts
+        uses: actions/upload-artifact@v3
+        with:
+          name: golden-test-updated
+          path: test/golden/goldens
+
+      - name: 💂 Verify that ${{ github.actor }} has permission to commit
+        if: ${{ github.event.inputs.commit == 'true' }}
+        shell: bash
+        run: git log --author='^${{ github.actor }} <\S*>$' --max-count=1 --exit-code > /dev/null && { echo '⚠ You have no commits on this branch and therefore no permission to commit ⚠' ; exit 1; } || exit 0
```
nt4f04uNd, line 36: This doesn't verify actual permissions - for example you don't have permissions to commit into any of the branches, but you have commits with your signature in the tree.

I would honestly prefer that we rather just limit this workflow to repository admins, so I as reviewer can trigger it on a pull request before I merge, after I reviewed all the goldens manually. It will regenerate all goldens and it's not ideal but I think it's an okay policy and would avoid potential flakes that could happen because of tolerance gap.

nt4f04uNd (commented):
This is a cool feature

.github/workflows/update_goldens.yml
nt4f04uNd: By admins I meant collaborators, actually

Abestanis: > This doesn't verify actual permissions

Yes, I mentioned this in the PR description briefly.This was the easiest way that I could think of that allows people to commit to their own branches or "collaborators" which have previously committed. I was hesitant to use the [actual collaborators list from the GitHub repository](https://github.com/nt4f04uNd/sweyer/graphs/contributors), because it is empty, since you squash all commits on a merge request, so GitHub counts the code as your contribution.

I will look into an alternative solution, because it would be nice if someone at least could commit into their own branches.

nt4f04uNd: > I was hesitant to use the [actual collaborators list from the GitHub repository](https://github.com/nt4f04uNd/sweyer/graphs/contributors), because it is empty, since you squash all commits on a merge request, so GitHub counts the code as your contribution.

Github doesn't count squashed commits as my contribution, since they are signed with the email of the pull request author, which is you. I might be co-author of the squshed commit, but that is only if I commited to pull request branch.

This list also doesn't display who is a collaborator on the repository, this is a list of all contributors.
Also, Github counts contibutions only on master branches of non-fork repositories, hence it doesn't display you in the list, until https://github.com/nt4f04uNd/sweyer/commits/1.0.8 is merged into master.

Abestanis: Ok, thanks for the clarification, I just assumed that the list would be empty based on what I could see. I would still like to try to allow people to commit to their own branches, its just very convenient. How about this:

We can check if someone is a contributor using [`actions-cool
/check-user-permission`](https://github.com/actions-cool/check-user-permission)
If the branch has a parent, we allow everyone who has made a commit on that branch up to it's parent, meaning we are looking just at the new commits on the branch.

Would that be ok with you?

Abestanis: Actually, I wonder if checking for the write user-permission is enough...

Abestanis: @nt4f04uNd could you please test if you can run the workflow with auto commit against `master` in [my fork](https://github.com/Abestanis/sweyer/actions/workflows/update_goldens.yml)? Also, try to open a pull request and check the auto commit against that branch.

Abestanis: Hang on, it's not working just yet.

Abestanis: Ok, you can try now

nt4f04uNd: > could you please test if you can run the workflow with auto commit against master in [my fork](https://github.com/Abestanis/sweyer/actions/workflows/update_goldens.yml)?

I cannot, there is no button to run a workflow for me.

I see that you try to disable only the step that actually commits goldens, but I would prefer to make this entire workflow only accessible to collaborators, it otherwise incentivizes to run it for no reason - reviewers should decide if new goldens are good or not.

nt4f04uNd: Ah, nevermind, workflow_dispatch is by default only accessible to collaborators

.github/workflows/update_goldens.yml
```diff
@@ -0,0 +1,41 @@
...
+
+      - name: 💂 Verify that ${{ github.actor }} has permission to commit
+        if: github.event.inputs.commit == 'true'
+        id: permissionCheck
+        uses: actions-cool/check-user-permission@v2
+        with:
+          require: write
+      - name: 💾 Commit updated Golden test artifacts
+        uses: stefanzweifel/git-auto-commit-action@v4
+        if: github.event.inputs.commit == 'true' && steps.permissionCheck.outputs.require-result == 'true'
```
nt4f04uNd, line 39: The workflow should fail if it doesn't have permissions to commit

So it's better to remove this entire thing witrh `permissionCheck`

nt4f04uNd (approved):
LGTM, except one comment

.github/workflows/update_goldens.yml
Abestanis: Ok, I think it is actually perfect that you can only run the workflow as a collaborator, because you can always run it on your own local fork. No need to run it here.

I will update the contributing documentation accordingly. Please merge this into `1.0.8` as well afterwards.

nt4f04uNd:
Oh I just realized that I would not be able to trigger this on forked branches, right? That would be better if I didn't have to run it on master... Any ideas?

Abestanis:
> Oh I just realized that I would not be able to trigger this on forked branches, right?

Yes, but the owner of the fork can.

PR #37, opened 2022-05-15
Don't fail the CI if coverage decreases

Abestanis:
This was enabled by default, now the two codecov status checks will always succeed.

This also fixes one source of unstable coverage, the `AudioHandler` wasn't properly re-initialized between tests. Let's see if that was the only source of coverage instability.

### DIFF
--- a/lib/logic/player/music_player.dart
+++ b/lib/logic/player/music_player.dart
@@ -658,7 +658,12 @@ class MusicPlayer extends AudioPlayer {
 
   Future<void> init() async {
     await restoreLastSong();
-    handler?._init(this);  // In case the handler already exists, reinitialize it.
+
+    // Reinitialize the AudioHandler if it already exists. Otherwise it is
+    // initialized by the AudioService. The AudioService must only ever be
+    // initialized once per process, but the handler depends on the MusicPlayer,
+    // which can be disposed and recreated.
+    handler?._init(this);
     handler ??= await AudioService.init(builder: () {
         return AudioHandler(MusicPlayer.instance);
       },
--- a/lib/logic/player/music_player.dart
+++ b/lib/logic/player/music_player.dart
@@ -180,16 +180,22 @@ class _BrowserParentProvider {
   }
 }
 
-class _AudioHandler extends BaseAudioHandler with SeekHandler, WidgetsBindingObserver {
-  _AudioHandler() {
-    _init();
+@visibleForTesting
+class AudioHandler extends BaseAudioHandler with SeekHandler, WidgetsBindingObserver {
+  AudioHandler(MusicPlayer player) {
+    _init(player);
   }
 
   bool _disposed = false;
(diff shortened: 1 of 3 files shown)


Abestanis:
If you run the `play/pause button works` in `player_route_test.dart` alone, it works, If you run all tests in `player_route_test.dart`, the test fails, because the `running` attribute of the AudioHandler is not `true`, because this callback is not called:
https://github.com/nt4f04uNd/sweyer/blob/e28bd86a97dd3444cd656a6799fc382aaec7ffa9/lib/logic/player/music_player.dart#L203-L208

@nt4f04uNd do you have any idea what's going on here? I verified that the `play` method of the audio player was called.

nt4f04uNd:
It's likely because the handler is initialized only once, statically.

lib/logic/player/music_player.dart
```diff
@@ -636,20 +645,22 @@ class MusicPlayer extends AudioPlayer {
...
+    if (song != null && handler!.running) {
+      handler!.mediaItem.add(song.toMediaItem());
     }
   }
 
   Future<void> init() async {
     await restoreLastSong();
-    _handler ??= await AudioService.init(builder: () {
-        return _AudioHandler();
+    await handler?._init(this);
```
nt4f04uNd, line 661: Can you leave a comment why do we need to do this here?

lib/logic/player/music_player.dart
```diff
@@ -180,16 +180,22 @@ class _BrowserParentProvider {
...
   bool _disposed = false;
-  bool _running = false;
-  MusicPlayer player = MusicPlayer.instance;
-
-  Future<void> _init() async {
+  @visibleForTesting
+  bool running = false;
+  late MusicPlayer player;
+  late StreamSubscription playbackSubscriber;
+  late StreamSubscription queueSubscriber;
```
nt4f04uNd, line 194: This will make calling dispose before init a crash, make those nullable

Abestanis: Is calling dispose before init really something that we want to allow? For me crashing there sounds like a good way to find bugs.

nt4f04uNd: It should be an OK action to call dispose in not initialized state

lib/logic/player/music_player.dart
```diff
@@ -658,7 +658,7 @@ class MusicPlayer extends AudioPlayer {
 
   Future<void> init() async {
     await restoreLastSong();
-    await handler?._init(this);
```
nt4f04uNd: Can you add more context to this comment? Also, please prefer putting comments above the lines that they refer to

lib/logic/player/music_player.dart
Abestanis: Actually, `_init` is called in the constructor, it is therefore guaranteed that these fields are initialized.

PR #39, opened 2022-05-18
Add issue templates

Abestanis:
This adds two issue templates, which should help in writing issues and also automatically label the issue.
This is how it will look:

Issue template selection screen
Notice that there is still an option to create a raw issue, if the templates don't fit. (The edit template options is only visible for admins.)

* Bug report:
  Bug report template
* Enhancement request:
  Enhancement request template

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug.yml
+++ b/.github/ISSUE_TEMPLATE/bug.yml
@@ -1,13 +1,10 @@
 name: Bug report
-description: Report any issues that you have found with the Sweyer app. Please check open issues first, in case it has already been reported.
+description: Report any issues that you have found with the app.
 labels: [bug]
 body:
   - type: markdown
     attributes:
-      value: |
-        Thank you for taking the time and filling out this bug report!
-
-        Please [check open issues](https://github.com/nt4f04uNd/sweyer/issues) first, in case it has already been reported.
+      value: Thank you for taking the time and filling out this bug report!
   - type: textarea
     id: reproduction-steps
     attributes:
@@ -21,15 +18,17 @@ body:
     validations:
       required: true
   - type: textarea
-    id: result
+    id: expected
     attributes:
-      label: Outcome
-      placeholder: Tell us what went wrong
-      value: |
-        #### What did you expect?
-
-        #### What happened instead?
-        
+      label: What did you expect?
+      placeholder: Tell us what you thought should happen.
+    validations:
+      required: true
+  - type: textarea
+    id: outcome
+    attributes:
+      label: What happened instead?
+      placeholder: Tell us what went actually happened.
     validations:
       required: true
   - type: input
@@ -44,7 +43,7 @@ body:
     id: android-version
     attributes:
       label: Android version
-      description: The Andorid version of your device. You can leave this empty if you are sure that it's not important, if you are unsure please specify. 
+      description: The Andorid version of your device. 
       placeholder: e.g. Android 12 or Android Snow Cone
     validations:
       required: false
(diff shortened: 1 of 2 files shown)


.github/ISSUE_TEMPLATE/bug.yml
```diff
@@ -0,0 +1,50 @@
...
+      value: |
+        1. Where are you starting? What can you see?
+        2. What do you click?
+        3. More steps…
+    validations:
+      required: true
+  - type: textarea
+    id: result
+    attributes:
+      label: Outcome
```
nt4f04uNd, line 26: Why is this one section and not two?

.github/ISSUE_TEMPLATE/bug.yml
```diff
@@ -0,0 +1,50 @@
...
+      label: Sweyer version
+      description: You can find the version information in App Menu -> Settings at the bottom. If you compiled the app yourself, please specify the commit here.
+      placeholder: e.g. 1.0.7
+    validations:
+      required: true
+  - type: input
+    id: android-version
+    attributes:
+      label: Android version
+      description: The Andorid version of your device. You can leave this empty if you are sure that it's not important, if you are unsure please specify.
```
nt4f04uNd, line 47: Let's remove "You can leave this empty if you are sure that it's not important, if you are unsure please specify." part since if we are adding it here, we would want people to add it here, and this is explanation is a bit specific. Leaving this as non-required is sufficient for people to undesrtand they can leave this empty.

.github/ISSUE_TEMPLATE/enhancement.yml
```diff
@@ -0,0 +1,22 @@
...
+    attributes:
+      label: Suggestion
+      description: Please feel welcome to include screenshots or sketches.
+      placeholder: Tell us what you would like to do!
+      value: |
+        #### What would you like to do?
+
+        #### Why would you like to do it?
+
+        #### How would you like to achieve it?
```
nt4f04uNd, line 19: Can this be a placeholder?

.github/ISSUE_TEMPLATE/enhancement.yml
```diff
@@ -0,0 +1,22 @@
+name: Enhancement request
+description: Do you have a suggestion or feature request?
+labels: [enhancement]
+body:
+  - type: markdown
+    attributes:
+      value: Thank you for taking the time to propose a new feature or make a suggestion.
```
nt4f04uNd, line 7: "Thank you for taking the time to make a suggestion!" would be better

.github/ISSUE_TEMPLATE/enhancement.yml
```diff
@@ -0,0 +1,22 @@
+name: Enhancement request
```
nt4f04uNd, line 1: "Feature request"

.github/ISSUE_TEMPLATE/enhancement.yml
```diff
@@ -0,0 +1,22 @@
...
+description: Do you have a suggestion or feature request?
```
nt4f04uNd, line 2: "Do you have a suggestion?"

.github/ISSUE_TEMPLATE/enhancement.yml
Abestanis: Should I split this into multiple text areas too?

nt4f04uNd: No

Abestanis: Ok, do you want me to replace the existing placeholders with these three lines or just the last?

nt4f04uNd: Actually, just remove the value here and make the placeholder "I would like to ..."

Abestanis: You can use the suggestion button to show me what you mean.

When you respond on GitHub, this one: Suggestion button

nt4f04uNd:
Thanks!

PR #48, opened 2022-05-20
Fix two TODOs and improve the tests

Abestanis:
This resolves two TODOs, whose respective linked issues have been resolved. The `useInheritedMediaQuery` caused some issues in the tests, because the screen size was not properly set. This lead me to discover that there was an issue with the current goldens, where the player route was partially visible in the home route.

I added a `Container` with an orange background around the `TrackShowcase` to demonstrate the problem:
| Before | After |
|--------|-------|
| tabs_route songs_tab dark - Before | tabs_route songs_tab dark - After |

The problem was that the app bar height defaults to `56`, where as the `TrackShowcase` is `64`. This wasn't a problem on a real device, because the `Scaffold` adds a `24` pixel padding for status bar on Android (`mediaQuery.padding.top`), but this padding is not applied in the test. From a comparison by eye, this change doesn't alter the height of the tab bar on an actual device.


test/routes/home_route_test.dart
```diff
@@ -50,6 +50,8 @@ void main() {
     await tester.runAppTest(() async {
       expect(Permissions.instance.granted, true);
       expect(find.byType(Home), findsOneWidget);
+      expect(tester.getRect(find.byType(TrackShowcase)).top,
+          tester.getRect(find.byType(App)).height, reason: 'Player route must be offscreen');
```
Abestanis, line 54: This test fails without the fix to the `AppBar` height. It should prevent regressions in the future.

test/test.dart
```diff
@@ -78,15 +78,19 @@ final ArtistCopyWith artistWith = _testArtist.copyWith;
 
 /// Default l10n delegate in tests.
 final l10n = AppLocalizationsEn();
-const kScreenSize = Size(kScreenWidth, kScreenHeight);
 const kScreenHeight = 800.0;
 const kScreenWidth = 450.0;
+const kScreenPixelRatio = 3.0;
```
Abestanis, line 83: This is the default which is used for the existing goldens.

test/test.dart
```diff
@@ -168,16 +172,7 @@ extension WidgetTesterExtension on WidgetTester {
...
-      Center(
-        child: MediaQuery(
-          data: MediaQueryData.fromWindow(window).copyWith(size: kScreenSize),
-          child: const App(
-            debugShowCheckedModeBanner: false,
-          ),
-        ),
-      ),
-    );
+    await pumpWidget(const App(debugShowCheckedModeBanner: false));
```
Abestanis, line 175: The `Center` and `MediaQuery` were unnecessary and were causing differences in behavior of the tests vs. the real app. For example: `MediaQuery.of` in the build method of `App` fails on a real device, but would succeed in the test.

lib/routes/home_route/player_route.dart
```diff
@@ -681,6 +682,7 @@ class _MainTabState extends State<_MainTab> {
         appBar: AppBar(
           elevation: 0.0,
           backgroundColor: Colors.transparent,
+          toolbarHeight: TrackPanel.height(mediaQuery.textScaleFactor) - mediaQuery.padding.top,
```
nt4f04uNd, line 685: This does change the height of the appbar, and it can be seen in generated goldens as well

test/routes/home_route_test.dart
nt4f04uNd: This should rather verify that the player route is actually offscreen

lib/routes/home_route/player_route.dart
Abestanis: True, this is setting the height so it doesn't get too small if the system status bar is not visible. I thought the idea was that the app bar would be the exact same height as the `TrackPanel`, since one fades over the other when opening the player route.
It shouldn't be a problem if the app bar is larger, but it must not be smaller. What do you think about this?
```suggestion
          toolbarHeight: max(
              TrackPanel.height(mediaQuery.textScaleFactor) - mediaQuery.padding.top,
              theme.appBarTheme.toolbarHeight ?? kToolbarHeight
          ),
```

test/routes/home_route_test.dart
Abestanis: I tried that, but
```dart
expect(find.byType(TrackShowcase), findsNothing, reason: 'Player route must be offscreen');
```
fails because the finder still finds the TrackShowcase (even though `skipOffstage` is `true`).

Do you have a better way to test that it's offscreen?

nt4f04uNd: Seems like getting its rect and expecting it to fail should do the thing

Abestanis: Actually no, `hitTestable` can't be used to filter it. Do you have another way to get the rect that fails, when the widget is offscreen?

nt4f04uNd: It's good to me like it is then, but reformat it to 3 separate lines by arguments, it would look cleaner

nt4f04uNd:
Can also use `TrackPanel.height(textScaleFactor)` instead of `kSongTileHeight` on this line?

https://github.com/nt4f04uNd/sweyer/blob/c118b60a84b4c82692c022a6f097b4c942e58de8/lib/routes/home_route/player_route.dart#L123

Abestanis:
> Can also use TrackPanel.height(textScaleFactor) instead of kSongTileHeight on this line?

Done.

Abestanis:
Hang on, something went wrong with the golden update.

lib/routes/home_route/player_route.dart
```diff
@@ -681,6 +684,10 @@ class _MainTabState extends State<_MainTab> {
         appBar: AppBar(
           elevation: 0.0,
           backgroundColor: Colors.transparent,
+          toolbarHeight: math.max(
+              TrackPanel.height(mediaQuery.textScaleFactor) - mediaQuery.padding.top,
```
nt4f04uNd, line 688: extra indent here

Abestanis: I miss auto formatting. 😅

PR #50, opened 2022-06-04
Hide the settings button if the settings route was already opened

### LINKED ISSUE #41: Gear button should be hidden when already in menu
The gear icon button in the lower right during a selection is used to enter a menu. Once in the menu, the button has no function and is a bit confusing. It should be hidden inside the menu:
Gear icon in menu issue

Abestanis:
Fixes #41.

Settings button behaviour


nt4f04uNd:
I don't think we need to update goldens here

lib/routes/selection_route.dart
```diff
@@ -16,28 +16,33 @@ class SelectionRoute extends StatefulWidget {
 class _SelectionRouteState extends State<SelectionRoute> {
   late final HomeRouter nestedHomeRouter = HomeRouter.selection(widget.selectionArguments);
   late final ContentSelectionController controller;
+  late bool settingsOpened;
 
   @override
   void initState() { 
     super.initState();
-    var settingsOpened = false;
```
nt4f04uNd, line 23: just set it to false in initializer and make it non-late

lib/routes/selection_route.dart
```diff
@@ -16,28 +16,33 @@ class SelectionRoute extends StatefulWidget {
...
-                  settingsOpened = true;
+                  setState(() {
+                    settingsOpened = true;
+                  });
                   await nestedHomeRouter.navigatorKey.currentState!.push(StackFadeRouteTransition(
                     child: Builder(builder: (context) => settingsPageBuilder(context)),
                     transitionSettings: AppRouter.instance.transitionSettings.greyDismissible,
                   ));
-                  settingsOpened = false;
+                  setState(() {
```
nt4f04uNd, line 43: we are calling it after `await` - put it inside

```dart
if (mounted) {

}
```

lib/routes/selection_route.dart
```diff
@@ -16,28 +16,33 @@ class SelectionRoute extends StatefulWidget {
...
-    var settingsOpened = false;
+    settingsOpened = false;
     controller = ContentSelectionController.createAlwaysInSelection(
       context: context,
       actionsBuilder: (context) {
         final l10n = getl10n(context);
         final settingsPageBuilder = widget.selectionArguments.settingsPageBuilder;
         return [
-          if (settingsPageBuilder != null)
+          if (settingsPageBuilder != null && !settingsOpened)
```
nt4f04uNd, line 31: maybe add a `SelectionAppBar.defaultSelectionActionsBuilder` animation, using AnimatedSwitcher? also set similar curves and duration to what we have for other AnimatedSwitcher animations

Abestanis:
> I don't think we need to update goldens here

I added a new golden test, the rest is just noise.

test/golden/routes_golden_test.dart
```diff
@@ -152,6 +152,16 @@ void main() {
         await tester.pumpAndSettle();
       }, goldenCaptureCallback: () => tester.screenMatchesGolden(tester, 'selection_route.selection_route'));
     });
+    testAppGoldens('selection_route_settings', (WidgetTester tester) async {
+      await tester.runAppTest(() async {
+        HomeRouter.instance.goto(HomeRoutes.factory.content<Playlist>(playlistWith()));
+        await tester.pumpAndSettle();
+        await tester.tap(find.byIcon(Icons.add_rounded).first);
+        await tester.pumpAndSettle();
+        await tester.tap(find.byIcon(Icons.settings_rounded).hitTestable().first);
```
nt4f04uNd, line 161: use `find.descendant` instead, so it is semantically clear what button we expect to be pressed

lib/routes/selection_route.dart
Abestanis: I'm not super familiar with the animation system, did you mean `AnimationSwitcher`?
Or is this what you had in mind?
```dart
if (settingsPageBuilder != null)
            AnimatedSwitcher(
              duration: const Duration(milliseconds: 240),
              switchInCurve: Curves.easeOutCubic,
              switchOutCurve: Curves.easeInCubic,
              transitionBuilder: SelectionAppBar.defaultSelectionActionsBuilder,
              child: settingsOpened
                ? const SizedBox.shrink()
                : NFIconButton(
                   ...
                ),
            ),
```

It doesn't show the config button when opening the selection route for the first time though. I would appreciate if you could give me some pointers.

nt4f04uNd: You are correct with your code snippet. The reason it doesn't work is actually a bug in defaultSelectionActionsBuilder, I pushed a fix to the nt4f04unds_widgets, please update the lock file to get this fix

nt4f04uNd: Actually, I don't like the looks of `SelectionAppBar.defaultSelectionActionsBuilder` here, let's just keep the defualt transitionBuilder (fade transition)

Abestanis:
Hm, looks like the goldens are still producing different images, even though I can't see a difference anymore. That's annoying.

test/golden/routes_golden_test.dart
```diff
@@ -152,6 +152,20 @@ void main() {
...
+        HomeRouter.instance.goto(HomeRoutes.factory.content<Playlist>(playlistWith()));
+        await tester.pumpAndSettle();
+        await tester.tap(find.byIcon(Icons.add_rounded).first);
+        await tester.pumpAndSettle();
+        await tester.tap(find.descendant(
+            of: find.byType(SelectionRoute),
+            matching: find.byIcon(Icons.settings_rounded),
+        ));
+        await tester.pumpAndSettle();
+      }, goldenCaptureCallback: () => tester.screenMatchesGolden(tester, 'selection_route.selection_settings'));
```
nt4f04uNd, line 167: please make the test name and `selection_route.<what_goes_here>` to match with each other, i.e. the test should be named `selection_settings`

nt4f04uNd:
> Hm, looks like the goldens are still producing different images, even though I can't see a difference anymore. That's annoying.

If you mean that goldens are different each time, yeah, this is annoying

test/golden/routes_golden_test.dart
nt4f04uNd: The naming of the golden files is `group.name`, this test would be the only one to not follow this rule.

test/golden/routes_golden_test.dart
```diff
@@ -152,6 +152,20 @@ void main() {
         await tester.pumpAndSettle();
       }, goldenCaptureCallback: () => tester.screenMatchesGolden(tester, 'selection_route.selection_route'));
     });
+
+    testAppGoldens('selection_route.selection_route_settings', (WidgetTester tester) async {
```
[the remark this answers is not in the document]
Abestanis: Ah, sorry, I was blind. I see now what you mean.

PR #51, opened 2022-06-04
Fix artist text being cut off at the top when too long

### LINKED ISSUE #44: The artist is cut up on the top in the queue
For long artist names, the top part of the text is cut of in the queue:

Cut of text

Abestanis:
This fixes all text with `subtitle2` style. Looks like we can't use a height of less than `1`. Fixes #44.

| Before | Song list item - Before |
|--------|--|
| **After**   | Song list item - After |


nt4f04uNd:
I think we don't need to regenerate goldens here, the change is very subtle

Abestanis:
> I think we don't need to regenerate goldens here, the change is very subtle

Yes, but there is change which creates noise when we want to check for an actual change later. We should keep a clean baseline, otherwise we'll introduce an unintentional change and not notice it in the noise.

nt4f04uNd:
> Yes, but there is change which creates noise when we want to check for an actual change later. We should keep a clean baseline, otherwise we'll introduce an unintentional change and not notice it in the noise.

I'm not seing any changes to the goldens at all. For such small patches to 1.0.8 branch let's update goldens only when the change will obviosly affect them.

nt4f04uNd:
Actually, I see them now, let's update them then.

Abestanis:
> Actually, I see them now, let's update them then.

Yes, that will reduce the noise on the other pull requests.
I would suggest either use the workflow to update them on the `1.0.8` branch or merge this one first, once the updated goldens have arrived.

PR #52, opened 2022-06-04
Make sure the trailing text in the ArtistWidget is always visible

Abestanis:
This ensures that the dot and the length of each song item are always visible and only the artist name is cut of with an ellipsis.

| Before | Song list item - Before |
|--------|--|
| **After**   | Song list item - After |


Abestanis:
It changes the space between the artist and the dot by a few pixels, (as shown by the goldens), but the change is minimal.

Issue #47, opened 2022-05-18
The media player notification stop button behaves unexpected


Abestanis:
<details>
<summary>The stop button in the Android media player notification shows some unexpected behavior.</summary>

![Notification stop button behaviour](https://user-images.githubusercontent.com/6966049/169153940-d04c3061-609b-493d-ba5c-e3d8d15696b5.gif)
</details>

There are two issues:
1. [x] Pressing on the stop button while the app is playing music does stop the playback, but the notification is not updated and keeps advancing the timer, as shown in the gif. The stop button only works as expected when the player is paused.
2. [x] The icon of the stop button is currently an `X`, which indicates that pressing it should close the notification. Instead, it only closes the media session and on newer Android version, the notification stays visible by default, when the media session ends. The user has to explicitly configure the notification to close, as shown in the gif. I propose that the icon should change to a square `⏹`, which would make the stopping intent clear and prevent the confusion.

PR #57, opened 2022-06-06
Fix not updating the player notification when stopping playback

Abestanis:
The `!running` check caused us not to update the state of the player notification when stopping the player. Fixes the first part of #47.

Notification stop button behavior


test/fakes/fake_just_audio.dart
```diff
@@ -73,6 +73,14 @@ final icyMetadataMessage = IcyMetadataMessage(
   ),
 );
 
+extension SaveCompleter<T> on Completer<T> {
```
nt4f04uNd, line 76: you can remove the name to make it non-public like this `extension on Completer<T>`

test/logic/player/music_player.dart
```diff
@@ -0,0 +1,109 @@
+import 'package:audio_service/audio_service.dart';
```
nt4f04uNd, line 1: please rename this file into `music_player_test.dart`

test/fakes/fake_just_audio.dart
Abestanis: Nice, didn't know that was possible.

test/logic/player/music_player.dart
Abestanis: 👍 Maybe rename `test/logic/player/queue.dart` as well for consistency.

PR #58, opened 2022-06-07
Fix the back button handling

### LINKED ISSUE #33: Dialogs don't have priority for back button events
When pressing the back button, the app currently prioritizes poping routes rather than the dialog, The dialog only reacts to the back button press when there is no route to pop, e.g. when the home route is reached.

<details>
<summary>Gifs showcasing the issue</summary>

Dialog back button Error 1
Dialog back button Error 2
Dialog back button Error 3
</details>

Abestanis:
The `BackButtonListener`s take precedence over generating a `Navigator.pop` that would close the dialog.
The solution to register our own back button listener when opening a dialog, which closes the dialog on a back button press.
This fix (32f0e0aab6519824defa25af9cc9e63ac1326d38) could go into nt4f04uNd/nt4f04unds_widgets#11. Fixes #33.

The above fix surfaced another problem with the back button listeners:
When the widget tree get's rebuild, they re-register their back button listeners and take priority ([source](https://github.com/flutter/flutter/blob/2aa348b9407e96ffe4eca8e8f213c7984afad3f7/packages/flutter/lib/src/widgets/router.dart#L1144-L1152)):
```dart
  @override
  void didUpdateWidget(covariant BackButtonListener oldWidget) {
    super.didUpdateWidget(oldWidget);
    if (oldWidget.onBackButtonPressed != widget.onBackButtonPressed) {
      dispatcher?.removeCallback(oldWidget.onBackButtonPressed);
      dispatcher?.addCallback(widget.onBackButtonPressed);
      dispatcher?.takePriority();
    }
  }
```

This caused our manually registered back button listener to move back in the priority list, even though it should be in the front. Rebuilding the widget tree was caused by showing/hiding the software keyboard in the `New playlist` dialog.
I added a `NFBackButtonListener` to address the problem. (I know, we just removed that. 😉) It adds a wrapper handler function so it doesn't have to re-register the listener when the listening function changes. I think this commit (cb7c2d37941bb86c83693621df8146dd7b22da3a) should also be include in nt4f04uNd/nt4f04unds_widgets#11.


nt4f04uNd:
> This caused our manually registered back button listener to move back in the priority list, even though it should be in the front.

Seems like this is a consequence of that you added a BackButtonListener in PlayerRoute (which we don't need, as I mentioned above), so removing it and just sticking to the default BackButtonListener from framework seems fine for now. I don't think that just removing `didUpdateWidget` is correct, neither is just removing `takePriority`, ideally you should report this to Flutter repository with a complete minimal reporducible sample (which I suspect is having BackButtonListeners in Stack).

nt4f04uNd:
So our issue here is actually https://github.com/flutter/flutter/issues/82046. This would work with our current code if BackButtonListener was muted when dialog route is pushed to the Stack.

lib/routes/routes.dart
```diff
@@ -520,30 +520,6 @@ class HomeRouter extends RouterDelegate<HomeRoutes<Object?>>
       !(_currentSearchDelegate?.chipsBarDragged ?? false);
   }
 
-  /// Callback that must be called before any pop.
-  /// 
-  /// For example we want that player route would be closed first.
-  bool handleNecessaryPop() {
```
nt4f04uNd, line 526: Refactoring it like you did breaks pop behavior for both of them on any route, except the starting one - search route, playlist route, artist route and there might be more.

You just had to move the whole `handleNecessaryPop` to the same place you moved the part of it - `TabsRouteState ._handlePop`, instead of splitting it into individual BackButtonListeners.

This overall is not needed for to fix the issue, but it is an OK refactor.

nt4f04uNd:
Pushed a workaround for the https://github.com/flutter/flutter/issues/82046 into https://github.com/nt4f04uNd/nt4f04unds_widgets/pull/11/commits/21012ef4b4c8e96d0ae420df0d36272ae7bbd33a

test/routes/player_route_test.dart
```diff
@@ -215,4 +216,34 @@ void main() {
     });
     expect(MusicPlayer.handler!.running, false);
   });
+
+  testWidgets('handles back presses correctly', (WidgetTester tester) async {
```
nt4f04uNd, line 220: This test passes on 1.0.8 without any changes.

nt4f04uNd: Actually, no, I had local changes on this branch, my bad

nt4f04uNd:
Thank you for investigating this, I applied some fixes I mentioned and merged.

PR #59, opened 2022-06-07
Add a padding between the song title and the favorite indicator

Abestanis:
| Before    | Before |
|-----------|-------|
| **After** | After |

### DIFF
--- a/lib/widgets/favorite_indicator.dart
+++ b/lib/widgets/favorite_indicator.dart
@@ -23,7 +23,7 @@ class FavoriteIndicator extends StatelessWidget {
       child: !shown
         ? const SizedBox.shrink()
         : Padding(
-            padding: const EdgeInsets.only(right: 8.0),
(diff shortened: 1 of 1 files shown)


nt4f04uNd:
Ok, this was an oversight on my part, but why is this only for song tile though? Can you just make the padding inside FavoriteIndicator symmetric?

Abestanis:
> Can you just make the padding inside FavoriteIndicator symmetric?

Done.

Abestanis:
I'm also regenerating the goldens.

Abestanis:
> I'm also regenerating the goldens.

Ok, looks like none of these differences were actually related to the changes in this PR, but I still think it would be good to update them so we get a good baseline again.

Abestanis:
> It's better to regenerate them to baseline where this is really needed to not clutter up the diff with lots of identical images.

Ok, we'll update them in #51.

PR #60, opened 2022-06-08
Make the golden tests consistent

Abestanis:
I know we said we should pause, but I think it's worth to merge this before you start on the update.

This makes the golden tests produce consistent output. I tested it by running the update goldens workflow three times in a row.
The main change is using `clock` everywhere and setting the clock during the tests to `binding.clock`, which is a consistent clock. I was surprised to see that the flutter test framework doesn't do this by default.


nt4f04uNd:
This is awesome! We could even set tolerance to 0.0, but let's keep it to allow local runs on different systems, and allow small differences in goldens between patches.

PR #62, opened 2022-06-16
Add editor configurations for our formatting line length

Abestanis:
This adds configuration for JetBrains editors and other editors respecting `.editorconfig` files to use the `120` character line length rule introduced in #61.

### DIFF
--- a/lib/routes/home_route/home_route.dart
+++ b/lib/routes/home_route/home_route.dart
@@ -136,8 +136,7 @@ class HomeState extends State<Home> {
       final now = clock.now();
       // Show toast when user presses back button on main route, that
       // asks from user to press again to confirm that he wants to quit the app
-      if (_lastBackPressTime == null ||
-          now.difference(_lastBackPressTime!) > constants.Config.backPressCloseTimeout) {
+      if (_lastBackPressTime == null || now.difference(_lastBackPressTime!) > constants.Config.backPressCloseTimeout) {
         _lastBackPressTime = now;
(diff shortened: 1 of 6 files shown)


lib/routes/home_route/home_route.dart
```diff
@@ -136,8 +136,7 @@ class HomeState extends State<Home> {
       final now = clock.now();
       // Show toast when user presses back button on main route, that
       // asks from user to press again to confirm that he wants to quit the app
-      if (_lastBackPressTime == null ||
```
Abestanis, line 139: Running `flutter format -l 120 .` formats this line. Is that an inconsistency between Flutter on Windows vs Linux or was it just missed for #61?

nt4f04uNd:
It seems to me that Android Studio adds its own rules to the formatting, beyond what `flutter format` does. I don't really want to encourage people to use its formatter, unless we have it do the same thing as barebone console command.

Abestanis:
True, Android Studio is annoyingly using it's own rules when you use the default formatting command and there is no way to disable that, but at least it moves the visual max line indicator in the editor. Also there is a `Reformat with "dart format"` command that does respect our settings.

I wouldn't say this encourages using the Android Studio formatter, this doesn't really change anything in that regard, it just improves the editing experience.

lib/routes/home_route/home_route.dart
nt4f04uNd: Seems like accidentally discarded this file

PR #63, opened 2022-06-16
Update dependencies

Abestanis:
This updates our dependencies, including some to a major version.

The `permission_handler` dependency can't be updated to `10.0.0`, because it's `permission_handler_android` dependency requires a `targetSdkVersion` of `33`. We are currently stuck at `30`, because we need an update of the `audio_service` dependency, as it creates a `PendingIntent` without specifying `FLAG_IMMUTABLE` which is an error from sdk version `31` onward.

The `flare` package is discontinued and replaced by `rive`. I had to manually recreate the animations in their editor, because it couldn't import the old files. I just created a temporary account for the rive editor which I don't intent to keep, so I'm giving you an export of the two files so hopefully you can import them (just remove the `txt` extension, GitHub doesn't allow me to upload them otherwise):
[play_pause.rev](https://github.com/nt4f04uNd/sweyer/files/8922165/play_pause.rev.txt)
[checkmark.rev](https://github.com/nt4f04uNd/sweyer/files/8922166/checkmark.rev.txt)
When you imported them, you can share them (try the `iframe` option) and put them [in the Readme](https://github.com/nt4f04uNd/sweyer/compare/1.0.8...Abestanis:feature/update_dependencies?expand=1#diff-93231595ea6823839ecd6586ba4b4d92b5a72d683aa19a32d87d3c64b7e3089dR3-R4).
https://github.com/nt4f04uNd/sweyer/blob/af3bde702b2afcb361a1ea06eef3bfd472e7d13f/assets/animations/README.md?plain=1#L3-L4
It turns out that `rive` is using a `Stopwatch` to drive it's animations, which doesn't respect the fake clock during test. This caused `pumpAndSettle` calls to time out. I have created a version of the package which uses a stopwatch from the `clock` package and proposed a pull request: https://github.com/rive-app/rive-flutter/pull/244
Once that is merged and a new version is released we can switch back to it.

The changes in the goldens are due to small differences in how the play button and selection indicators are rendered, but it's not a change that is visible to the eye. I also had to make the cursor blinking behavior consistent, see 2fcbffd8a0b5a51cced39e7e4853f33fb42fe4a4 and [the documentation](https://api.flutter.dev/flutter/widgets/EditableText/debugDeterministicCursor.html).

This gets rid of almost all of the `Operand of null-aware operation '!' has type 'WidgetsBinding' which excludes null.` warnings during build. The only ones remaining are from `nt4f04unds_widgets`, which I'll let you handle. I also didn't look into updating our other two custom dependencies (`fluttertoast`, `audio_service`). I would also leave that up to you.

### DIFF
--- a/assets/animations/README.md
+++ b/assets/animations/README.md
@@ -1,4 +1,16 @@
 ## Animation sources
 
-* [`checkmark.json`](checkmark.json)
-* [`play_pause.json`](play_pause.json)
+* [`checkmark.json`](https://lottiefiles.com/112787-check-mark)
+* [`play_pause.json`](https://lottiefiles.com/112788-play-pause)
+
+Source Adobe After Effects projects for these animations can also be found by those links.
+
+## How to create animations
+
+These animations are created and run by Lottie.
+
+To edit these animations Adobe After Effects can be used, together with:
+
+* [Bodymovin plugin](https://github.com/airbnb/lottie-web#plugin-installation) to import
+the files
+* and [Lottie Files plugin](https://lottiefiles.com/plugins/after-effects) to export them


test/test.dart
```diff
@@ -249,6 +246,7 @@ void testAppGoldens(
   bool? skip,
   Object? tags = _defaultTagObject,
 }) {
+  EditableText.debugDeterministicCursor = true;
```
nt4f04uNd, line 249: This should copy the current value of it and apply it after the test ends (use try/finally for that)

nt4f04uNd:
Rive is bad, to be honest. They are now limiting the amount of files you can use for free - once you reach the limit, you can't use their format, because the editor is behind a pay-wall. They removed the community hub.

I'd really want to move to the lottie, but it's another time investment. I'd want to stick with the flare until it breaks us entirely. Warnings are OK to be there, they are just warnings.

Abestanis:
> Warnings are OK to be there, they are just warnings.

Yeah, but they are annoying and distracting from warnings which actually indicate a real problem. I would prefer if we could switch to rive for now and move to lottie later, since I already put in the time to migrate. At least with this we don't have the warnings anymore.

nt4f04uNd:
Regardning `audio_service`, see https://github.com/nt4f04uNd/sweyer/issues/76

Abestanis:
> Regardning audio_service, see #76

Hm, are we already behind so much that we cant rebase on the latest `master`?

nt4f04uNd:
> Yeah, but they are annoying and distracting from warnings which actually indicate a real problem. I would prefer if we could switch to rive for now and move to lottie later, since I already put in the time to migrate. At least with this we don't have the warnings anymore.

I appreciate the effort, but I don't want to convert them into rive.

Just things that I immediately noticed:

* the console is being spammed with this when `play_pause` animation is in pause state
```
D/CCodecBufferChannel(17912): [c2.android.aac.decoder#132] DEBUG: elapsed: n=6 [in=0 pipeline=0 out=2]
D/PipelineWatcher(17912): DEBUG: elapsed 2 / 6
I/CCodecBufferChannel(17912): pipelineRoom(2) is <= numClientBuffers(2). Not signalling any more buffers to client
I/CCodecBufferChannel(17912): pipelineRoom(1) is <= numClientBuffers(1). Not signalling any more buffers to client
I/CCodecBufferChannel(17912): pipelineRoom(2) is <= numClientBuffers(2). Not signalling any more buffers to client
I/CCodecBufferChannel(17912): pipelineRoom(1) is <= numClientBuffers(2). Not signalling any more buffers to client
I/CCodecBufferChannel(17912): pipelineRoom(1) is <= numClientBuffers(1). Not signalling any more buffers to client
I/CCodecBufferChannel(17912): pipelineRoom(1) is <= numClientBuffers(2). Not signalling any more buffers to client
I/CCodecBufferChannel(17912): pipelineRoom(1) is <= numClientBuffers(1). Not signalling any more buffers to client
2
I/CCodecBufferChannel(17912): pipelineRoom(2) is <= numClientBuffers(2). Not signalling any more buffers to client
I/CCodecBufferChannel(17912): pipelineRoom(1) is <= numClientBuffers(2). Not signalling any more buffers to client
I/CCodecBufferChannel(17912): pipelineRoom(1) is <= numClientBuffers(1). Not signalling any more buffers to client
I/System.out(17912): ### AudioHandlerInterface message: setState
I/CCodecBufferChannel(17912): pipelineRoom(1) is <= numClientBuffers(1). Not signalling any more buffers to client
I/CCodecBufferChannel(17912): pipelineRoom(1) is <= numClientBuffers(2). Not signalling any more buffers to client
I/CCodecBufferChannel(17912): pipelineRoom(1) is <= numClientBuffers(1). Not signalling any more buffers to client
I/CCodecBufferChannel(17912): pipelineRoom(2) is <= numClientBuffers(2). Not signalling any more buffers to client
I/CCodecBufferChannel(17912): pipelineRoom(2) is <= numClientBuffers(3). Not signalling any more buffers to client
I/CCodecBufferChannel(17912): pipelineRoom(2) is <= numClientBuffers(2). Not signalling any more buffers to client
```
* I'm seing the same artifacts that I reported here and thankfully was able to workaround with flare https://github.com/2d-inc/Flare-Flutter/issues/286
* Animations are not accurate to what we currently have, because you converted them by hands

Taking all of this into account, I don't think it's worth the effort to migrate onto rive.

To prevent such things from happening in the future, please open issues to discuss whether it's actually worth to invest time in fixing some issue, and whether it's actually an issue, because we only have a warning that we can ignore right now. It doesn't indicate any real issue.

nt4f04uNd:
Sorry, I linked a wrong issue https://github.com/2d-inc/Flare-Flutter/issues/253#issuecomment-690336113

nt4f04uNd:
> Animations are not accurate to what we currently have, because you converted them by hands

Actually no, they are pretty accurate, but the other two issues need to be fixed before I can merge this.

nt4f04uNd:
First of all, thank you, this is really and outstanding contribution!

> I used Adobe After Effects to create the lottie animations, if you want I can provide the project file.

You mentioned that you were able to "convert the rive checksum animation to lottie". What does that mean and what are the source assets that you imported into this PR?

Can we upload the source files to the lottie website? We need to have some sort of source file that we can edit and from which we can re-export those files.

lib/widgets/selection.dart
```diff
@@ -743,13 +743,13 @@ class SelectionCheckmark extends StatefulWidget {
   _SelectionCheckmarkState createState() => _SelectionCheckmarkState();
 }
 
-class _SelectionCheckmarkState extends State<SelectionCheckmark> {
-  String _flareAnimation = 'stop';
+class _SelectionCheckmarkState extends State<SelectionCheckmark> with TickerProviderStateMixin {
```
nt4f04uNd, line 746: SingleTickerProviderStateMixin

Abestanis:
> You mentioned that you were able to "convert the rive checksum animation to lottie". What does that mean ...

Ups, I meant to say `checkmark` instead of `checksum`. I was able to use [this software](https://github.com/rive-app/rive-lottie) to convert the `.riv` file into a lottie `.json` file. This process only worked for the `checkmark.riv`, not for the `play_pause.riv`.

> ... what are the assets that you imported into this PR?

For the `checkmark`, I imported the `.json` into After Effects using the [Bodymovin plugin](https://github.com/airbnb/lottie-web#plugin-installation), cleaned up the animation (removed some artifacts generated by the conversion), added some proper names to the layer and shape, and exported the animation using the [Lottie Files plugin](https://lottiefiles.com/plugins/after-effects). The generated json is what I have included into this PR. For the `play_pause` animation the I created the animation in After Effects from scratch and then exported it the same way.

> Can we upload the source files to the lottie website?

Yes, could you create an account and upload them please? Afterwards you can update the [README](https://github.com/nt4f04uNd/sweyer/blob/1b1756a5f47261618251e867d87c000b58e9e740/assets/animations/README.md?plain=1#L3-L4) with the new links.

> We need to have some sort of source file that we can edit and from which we can re-export those files.

I can give you the After Effects project file, but all we should need are the `json` files that I included in the PR. You can import them into After Efffects using the [Bodymovin plugin](https://github.com/airbnb/lottie-web#plugin-installation).

lib/widgets/selection.dart
Abestanis: Cool, I didn't know about that one before.

Abestanis:
> Would you mind sharing after effects files into a comment here, just so I publish them together with lottie files?

Sure (had to put them into a zip because GitHub doesn't like the file extension):
[animations.zip](https://github.com/nt4f04uNd/sweyer/files/9106849/animations.zip)

> Also, do I need to attribute to you somehow in description of those files maybe?

No, that is not necessary. I only copied the animations that you created.

nt4f04uNd:
Amazing

PR #65, opened 2022-06-17
Fix a bunch of typos

Abestanis:
This fixes a bunch small spelling mistakes in the codebase. This shouldn't introduce any visible changes in the app, except for one fixed typo in the English translation.


lib/localization/arb/intl_en.arb
```diff
@@ -88,7 +88,7 @@
   "allTracks": "All tracks",
   "allAlbums": "All albums",
   "allPlaylists": "All playlists",
-  "allArtists": "All artitst",
+  "allArtists": "All artists",
```
Abestanis, line 91: This should be the only user facing change.

PR #67, opened 2022-06-17
Fix layout issues caused by not accounting for the textScaleFactor

### LINKED ISSUE #66: Layout issues when increasing the system font size
### Steps to reproduce

1. Increase the system font size in the Android settings. (Anything that causes a `MediaQuery.textScaleFactor >= 1.1`)
2. Open the app.

### What did you expect?

There should be no layout issues.

### What happened instead?

The height given for the `SongTitle` widget is not enough to contain the two lines of text. This is because `kSongTileHeight` doesn't take the `textScaleFactor` into account.

Additionally, the tab header widget on the bottom is partially hidden behind the bottom sheet.

Layout issues on higher textscale settings

For context, the default font size setting on my phone produces a `textScaleFactor` of `1.1`. The old version of Sweyer (`1.0.7`) does not have this problem.

### Sweyer version

Latest on the `1.0.8` branch; 4460b5a9688edcfd4ac32e3f3530c22d0d0cdca9

### Android version

_No response_

Abestanis:
This fixes the layout issues seen when the system font size is increased. The `math.max` makes sure that the `kSongTileHeight` doesn't get too small. This is not the most elegant solution, since it requires us to pass the `textScaleFactor` around a lot, let me know if you find a better solution. Fixes #66.

| Largest font | Smallest font |
|--------------|----------------|
| Largest font | Smallest font |

### DIFF
--- a/lib/widgets/content_list_view/song_tile.dart
+++ b/lib/widgets/content_list_view/song_tile.dart
@@ -17,13 +17,12 @@ TextStyle _titleTheme(ThemeData theme) => theme.textTheme.headline6!;
 
 /// Calculate the height of one line of text rendered with the [style] and [textScaleFactor].
 double _lineHeight(TextStyle style, double textScaleFactor) {
-  final TextPainter textPainter = TextPainter(
+  return TextPainter(
     text: TextSpan(text: "", style: style),
     maxLines: 1,
     textDirection: TextDirection.ltr,
     textScaleFactor: textScaleFactor,
-  )..layout(minWidth: 0, maxWidth: double.infinity);
-  return textPainter.size.height;
+  ).preferredLineHeight;
 }
 
 /// The height of the title and subtitle part of the [SongTile] widget for the given [context].
--- a/lib/widgets/content_list_view/song_tile.dart
+++ b/lib/widgets/content_list_view/song_tile.dart
@@ -5,8 +5,30 @@ import 'package:sweyer/sweyer.dart';
 
 /// Needed for scrollbar label computations
 const double _tileVerticalPadding = 8.0;
+
+/// Calculate the height of one line of text rendered with the [style] and [textScaleFactor].
+double _lineHeight(TextStyle style, double textScaleFactor) {
+  final TextPainter textPainter = TextPainter(
+    text: TextSpan(text: "", style: style),
+    maxLines: 1,
+    textDirection: TextDirection.ltr,
+    textScaleFactor: textScaleFactor,
+  )..layout(minWidth: 0, maxWidth: double.infinity);
+  return textPainter.size.height;
+}
+
+/// The height of the title and subtitle part of the [SongTile] widget for the given [context].
+double kSongTileTextHeight(context) {
+  final textScaleFactor = MediaQuery.of(context).textScaleFactor;
+  final theme = Theme.of(context);
+  return _lineHeight(_SongTileState.titleTheme(theme), textScaleFactor) +
+      _lineHeight(ArtistWidget.defaultTextStyle(theme), textScaleFactor) +
+      _SongTileState.subtitleTopPadding +
+      _SongTileState.subtitleBottomPadding;
+}
+
 double kSongTileHeight(BuildContext context) =>
-    (kSongTileArtSize + _tileVerticalPadding * 2) * math.max(0.95, MediaQuery.of(context).textScaleFactor);
+    math.max(kSongTileArtSize, kSongTileTextHeight(context)) + _tileVerticalPadding * 2;
 const double kSongTileHorizontalPadding = 10.0;
 const SongTileClickBehavior kSongTileClickBehavior = SongTileClickBehavior.play;
 const SongTileVariant kSongTileVariant = SongTileVariant.albumArt;
@@ -165,6 +187,12 @@ class SongTile extends SelectableWidget<SelectionEntry> {
 
 class _SongTileState extends SelectableState<SelectionEntry<Song>, SongTile> with ContentTileComponentsMixin {
   Color? previousBackgroundColor;
+  /// The padding that is added to the top of the subtitle widget.
+  static const subtitleTopPadding = 4.0;
+  /// The padding that is added to the bottom of the subtitle widget.
+  static const subtitleBottomPadding = 3.0;
+  /// The [TextStyle] used for the title text from the [theme].
+  static TextStyle titleTheme(ThemeData theme) => theme.textTheme.headline6!;
 
   @override
   void didUpdateWidget(SongTile oldWidget) {
@@ -219,7 +247,7 @@ class _SongTileState extends SelectableState<SelectionEntry<Song>, SongTile> wit
     Widget title = Text(
       widget.song.title,
       overflow: TextOverflow.ellipsis,
-      style: theme.textTheme.headline6,
+      style: titleTheme(theme),
     );
     Widget subtitle = ArtistWidget(
       artist: widget.song.artist,
@@ -277,9 +305,9 @@ class _SongTileState extends SelectableState<SelectionEntry<Song>, SongTile> wit
                       mainAxisAlignment: MainAxisAlignment.center,
                       children: [
                         title,
-                        const SizedBox(height: 4.0),
+                        const SizedBox(height: subtitleTopPadding),
                         subtitle,
-                        const SizedBox(height: 3.0),
+                        const SizedBox(height: subtitleBottomPadding),
                       ],
                     ),
                   ),
--- a/lib/widgets/content_list_view/song_tile.dart
+++ b/lib/widgets/content_list_view/song_tile.dart
@@ -5,8 +5,8 @@ import 'package:sweyer/sweyer.dart';
 
 /// Needed for scrollbar label computations
 const double _tileVerticalPadding = 8.0;
-double kSongTileHeight(double textScaleFactor) =>
-    (kSongTileArtSize + _tileVerticalPadding * 2) * math.max(0.95, textScaleFactor);
+double kSongTileHeight(BuildContext context) =>
+    (kSongTileArtSize + _tileVerticalPadding * 2) * math.max(0.95, MediaQuery.of(context).textScaleFactor);
 const double kSongTileHorizontalPadding = 10.0;
 const SongTileClickBehavior kSongTileClickBehavior = SongTileClickBehavior.play;
 const SongTileVariant kSongTileVariant = SongTileVariant.albumArt;
--- a/lib/widgets/content_list_view/song_tile.dart
+++ b/lib/widgets/content_list_view/song_tile.dart
@@ -1,9 +1,12 @@
+import 'dart:math' as math;
+
 import 'package:flutter/material.dart';
 import 'package:sweyer/sweyer.dart';
 
 /// Needed for scrollbar label computations
-const double kSongTileHeight = kSongTileArtSize + _tileVerticalPadding * 2;
 const double _tileVerticalPadding = 8.0;
+double kSongTileHeight(double textScaleFactor) =>
+    (kSongTileArtSize + _tileVerticalPadding * 2) * math.max(0.95, textScaleFactor);
 const double kSongTileHorizontalPadding = 10.0;
 const SongTileClickBehavior kSongTileClickBehavior = SongTileClickBehavior.play;
 const SongTileVariant kSongTileVariant = SongTileVariant.albumArt;
--- a/lib/widgets/content_list_view/song_tile.dart
+++ b/lib/widgets/content_list_view/song_tile.dart
@@ -6,6 +6,15 @@ import 'package:sweyer/sweyer.dart';
 /// Needed for scrollbar label computations
 const double _tileVerticalPadding = 8.0;
 
+/// The padding that is added to the top of the subtitle widget.
+const _subtitleTopPadding = 4.0;
+
+/// The padding that is added to the bottom of the subtitle widget.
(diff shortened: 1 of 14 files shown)


lib/widgets/content_list_view/song_tile.dart
```diff
@@ -1,9 +1,12 @@
+import 'dart:math' as math;
+
 import 'package:flutter/material.dart';
 import 'package:sweyer/sweyer.dart';
 
 /// Needed for scrollbar label computations
-const double kSongTileHeight = kSongTileArtSize + _tileVerticalPadding * 2;
 const double _tileVerticalPadding = 8.0;
+double kSongTileHeight(double textScaleFactor) =>
+    (kSongTileArtSize + _tileVerticalPadding * 2) * math.max(0.95, textScaleFactor);
```
nt4f04uNd, line 9: This distorts the proportions of the art. Scale factor shouldn't ever decrease the size of the tile at all.

nt4f04uNd: I think it's better to pass BuildContext instead of double

Abestanis: > This distorts the proportions of the art. [...] at some point it surpasses the size of Album and Artist tiles. It also breaks selection indicators.

Hm, I'm unable to replicate any of this. The emulator only allows for a maximum text scale factor of `1.3`, maybe that's not enough.

But I agree, the current formula felt a bit like a hack. Just thinking out loud, the correct solution would be something like
```dart
kSongTileHeight(context) => math.max(kSongTileArtSize, kSongTitleTextSize(context)) + _tileVerticalPadding * 2
```
where `kSongTitleTextSize` would be the height of the song title text with respect to text scaling + the height of the artist name text with respect to the scaling + the padding between the two.

Do you agree?

Abestanis: I think dd1c4792a2c3ce934ef292c36faeff627eea6de9 is the correct way to do it. I verified with the layout inspector that we are calculating the exact resulting height of the SongTile.

Abestanis:
I can't tell if this introduces some visual changes in the default configuration. Please update the goldens on the `1.0.8` branch before merging this, so we can compare against a clean baseline again.

lib/widgets/content_list_view/song_tile.dart
```diff
@@ -1,9 +1,34 @@
...
+
+/// Calculate the height of one line of text rendered with the [style] and [textScaleFactor].
+double _lineHeight(TextStyle style, double textScaleFactor) {
+  final TextPainter textPainter = TextPainter(
+    text: TextSpan(text: "", style: style),
+    maxLines: 1,
+    textDirection: TextDirection.ltr,
+    textScaleFactor: textScaleFactor,
+  )..layout(minWidth: 0, maxWidth: double.infinity);
+  return textPainter.size.height;
```
nt4f04uNd, line 17: This is extremely expensive. Why do we even need a layout for that? Can't we just take the `fontSize * textScaleFactor` and that would be it?

lib/widgets/content_list_view/song_tile.dart
```diff
@@ -161,6 +186,12 @@ class SongTile extends SelectableWidget<SelectionEntry> {
 
 class _SongTileState extends SelectableState<SelectionEntry<Song>, SongTile> with ContentTileComponentsMixin {
   Color? previousBackgroundColor;
+  /// The padding that is added to the top of the subtitle widget.
+  static const subtitleTopPadding = 4.0;
+  /// The padding that is added to the bottom of the subtitle widget.
+  static const subtitleBottomPadding = 3.0;
+  /// The [TextStyle] used for the title text from the [theme].
+  static TextStyle titleTheme(ThemeData theme) => theme.textTheme.headline6!;
```
nt4f04uNd, line 194: I'd prefer these to be defined privately in global scope at the top of the file, like other constants

lib/widgets/content_list_view/song_tile.dart
nt4f04uNd: > Hm, I'm unable to replicate any of this. The emulator only allows for a maximum text scale factor of 1.3, maybe that's not enough.

The problem is that even with that value it scales too much unncesessarily. All this empty space is not supposed to be there.

<details>
<summary>screenshot</summary>

Screenshot_1656788299
</details>

nt4f04uNd: On my OPPO phone device I'm able to get it up to 1.6

Abestanis: > The problem is that even with that value it scales too much unncesessarily. All this empty space is not supposed to be there.

Where is this empty space coming from? We should be scaling with the text, but the padding should remain constant. I'll investigate...

lib/widgets/content_list_view/song_tile.dart
Abestanis: Unfortunately not, the values are off:
```dart
print('title height: painter=${_lineHeight(_titleTheme(theme), textScaleFactor)}, fontSize*textScale=${_titleTheme(theme).fontSize! * textScaleFactor}');
  print('ArtistWidget height: painter=${_lineHeight(ArtistWidget.defaultTextStyle(theme), textScaleFactor)}, fontSize*textScale=${ArtistWidget.defaultTextStyle(theme).fontSize! * textScaleFactor}');
```
produces
```
title height: painter=24.0, fontSize*textScale=15.0
ArtistWidget height: painter=13.0, fontSize*textScale=13.5
```
Looks like there is something else that goes into these calculations.

lib/widgets/content_list_view/song_tile.dart
Abestanis: Done: c183ede651696b4e3b449316f0ef9d8f01ae9889

lib/widgets/content_list_view/song_tile.dart
Abestanis: I found a better solution that doesn't require a full call to layout (77cb463ac02e313812323d05072f5791576e929a).
If you are still concerned about performance, we can cache the result using the [`memoize` package](https://pub.dev/packages/memoize).

lib/widgets/content_list_view/song_tile.dart
Abestanis: Do you still see a problem here with the latest changes? Honestly, I don't see the empty space you are talking about in your screenshot. Could you please clarify where the unnecessary space is?

nt4f04uNd: I was talking about the spacing between the title and subtitle, and that it increases. It comes from the text line height scaling. I wanted to lower this a bit, but now looking at this, I think this is fine.

nt4f04uNd:
> I found a better solution that doesn't require a full call to layout ([77cb463](https://github.com/nt4f04uNd/sweyer/commit/77cb463ac02e313812323d05072f5791576e929a)).
> If you are still concerned about performance, we can cache the result using the [`memoize` package](https://pub.dev/packages/memoize).

Yep, so

I actually tried writing an optimization for it and compared it with `preferredLineHeight` performance and accuracy.

### Accuracy

It's acceptable in accuracy, and such print outputs the following:
(But I still don't like it, I wrote why at the end)

<details>

```
final textScaleFactor = MediaQuery.of(context).textScaleFactor;
  print('fff textScaleFactor: $textScaleFactor, preferredLineHeight: ${math.max(kSongTileArtSize, kSongTileTextHeight(context)) + _tileVerticalPadding * 2}, optimized ${kSongTileArtSize + _tileVerticalPadding * 2 + _calculateTextHeight(context)}');

fff textScaleFactor: 1.0, preferredLineHeight: 64.0, optimized 64.0
fff textScaleFactor: 1.25, preferredLineHeight: 70.0, optimized 72.875
fff textScaleFactor: 1.5, preferredLineHeight: 79.0, optimized 81.75
fff textScaleFactor: 1.75, preferredLineHeight: 90.0, optimized 90.625
fff textScaleFactor: 2.0, preferredLineHeight: 99.0, optimized 99.5
fff textScaleFactor: 2.25, preferredLineHeight: 108.0, optimized 108.375
fff textScaleFactor: 2.5, preferredLineHeight: 118.0, optimized 117.25 <- optimized is less than preferredLineHeight
```
</details>

### Benchmark

<details>
<summary>Code</summary>

```dart
T measure<T>(Function callback) {
  final s = Stopwatch();
  s.start();
  final result = callback();
  s.stop();
  debugPrint('elapsed ${s.elapsedMicroseconds}');
  return result;
}

double kSongTileHeight(BuildContext context) {
  return measure<double>(() => _kSongTileHeight(context));
}

double _kSongTileHeight(BuildContext context) {
  return math.max(kSongTileArtSize, kSongTileTextHeight(context)) + _tileVerticalPadding * 2;

  // my optimization
  return kSongTileArtSize + _tileVerticalPadding * 2 + _calculateTextHeight(context);
}
```
</details>

<details>
<summary>preferredLineHeight</summary>

```
I/flutter (24925): elapsed 8291
I/flutter (24925): elapsed 241
I/flutter (24925): elapsed 256
I/flutter (24925): elapsed 320
I/flutter (24925): elapsed 302
I/flutter (24925): elapsed 1068
I/flutter (24925): elapsed 269
I/flutter (24925): elapsed 258
I/flutter (24925): elapsed 193
I/flutter (24925): elapsed 1207
I/flutter (24925): elapsed 541
I/flutter (24925): elapsed 220
I/flutter (24925): elapsed 312
I/flutter (25123): elapsed 286
I/flutter (25123): elapsed 277
I/flutter (25123): elapsed 316
I/flutter (25123): elapsed 534
I/flutter (25123): elapsed 330
I/flutter (25123): elapsed 275
I/flutter (25123): elapsed 302
I/flutter (25123): elapsed 239
I/flutter (25123): elapsed 267
I/flutter (25123): elapsed 275
I/flutter (25123): elapsed 286
I/flutter (25123): elapsed 310
I/flutter (25123): elapsed 398
I/flutter (25123): elapsed 1372
I/flutter (25123): elapsed 462
I/flutter (25123): elapsed 394
I/flutter (25123): elapsed 305
I/flutter (25123): elapsed 420
I/flutter (25123): elapsed 392
I/flutter (25123): elapsed 355
```
</details>

<details>
<summary>my optimization</summary>

```
I/flutter (25443): elapsed 93
2
I/flutter (25443): elapsed 4
I/flutter (25443): elapsed 5
I/flutter (25443): elapsed 14
I/flutter (25443): elapsed 32
I/flutter (25443): elapsed 10
I/flutter (25443): elapsed 5
I/flutter (25443): elapsed 12
I/flutter (25443): elapsed 38
I/flutter (25443): elapsed 14
2
I/flutter (25443): elapsed 5
3
I/flutter (25783): elapsed 21
2
I/flutter (25783): elapsed 17
I/flutter (25783): elapsed 14
I/flutter (25783): elapsed 17
I/flutter (25783): elapsed 14
I/flutter (25783): elapsed 16
I/flutter (25783): elapsed 25
I/flutter (25783): elapsed 18
I/flutter (25783): elapsed 17
I/flutter (25783): elapsed 19
I/flutter (25783): elapsed 15
I/flutter (25783): elapsed 18
I/flutter (25783): elapsed 25
I/flutter (25783): elapsed 27
```
</details>


<details>
<summary>memoization</summary>

```
I/flutter (19906): elapsed 28
2
I/flutter (19906): elapsed 23
I/flutter (19906): elapsed 15
I/flutter (19906): elapsed 17
I/flutter (19906): elapsed 21
I/flutter (19906): elapsed 27
I/flutter (19906): elapsed 25
I/flutter (19906): elapsed 29
I/flutter (19906): elapsed 22
I/flutter (19906): elapsed 28
I/flutter (19906): elapsed 18
I/flutter (19906): elapsed 16
I/flutter (19906): elapsed 21
I/flutter (19906): elapsed 31
I/flutter (19906): elapsed 18
I/flutter (19906): elapsed 24
I/flutter (19906): elapsed 32
I/flutter (19906): elapsed 20
I/flutter (19906): elapsed 21
2
I/flutter (19906): elapsed 20
I/flutter (19906): elapsed 21
I/flutter (19906): elapsed 486
I/flutter (19906): elapsed 25
I/flutter (19906): elapsed 31
I/flutter (19906): elapsed 22
I/flutter (19906): elapsed 19
I/flutter (19906): elapsed 27
I/flutter (19906): elapsed 30
```
</details>

### Discussion

One think I don't like about my optimization is that it's not accurate. Eventually, with values >= 2.5 (which is unrealistic, practically speaking, but possible in theory), it diverges more and more from `preferredLineHeight`.

As you suggested, I would like to use memoization here, so we keep the accuracy `preferredLineHeight` of and fairly optimize it.

The need in memoization is especially exacerbated by the fact that this method used to be a constant and, for example, its value was used in the animation of the player route (see `TrackPanel.height(context)`) - so it can be called on each frame. Without memoization we are forcing the app to do a lot fo unnecessary work.

nt4f04uNd:
I am merging this PR as-as and will implement memoization and further fixes to **textScaleFactor** in a separate PR myself.

Thank you for contributing to this project! ❤️

Abestanis:
Perfect, thank you for merging this.

I agree with you, caching definetly sounds like a good idea and the optimization, even though it would be nice and fast, will be a problem with larger font sizes.

It also potentially might be a problem with different fonts. Although I don't have an example of a specific font, TrueType fonts can have different appearances for different sizes and resolutions, so they might change height in a way that the optimization can't predict. (From [Microsoft's TrueType fundementals](https://learn.microsoft.com/en-us/typography/opentype/spec/ttch01): `TrueType instruction set provides a powerful means of controlling glyph outlines at all or specific sizes and resolutions.`)

PR #70, opened 2022-06-21
Refactor the content type system

Abestanis:
This is a refactor of the content type system. See https://github.com/nt4f04uNd/sweyer/pull/34#discussion_r878897458 for more details.

The main problem that this is trying to solve is to move the representation of a content type from the current unbound representation `Type` to an enum. In Dart, `Type` can't be restricted to certain classes and even if we could specify a required common parent class, it would still allow to pass abstract intermediate classes, such as `SongOrigin`. An enum is meant to be used for this situation, a type that has only a limited number of values.

Benefits:
* We can get rid of a bunch of asserts that a `Type` must be a subclass of `Content`, because it's now impossible to pass anything but a content type to a function which expects one.
* We can get rid of `contentPick`, because we can use a `switch` statement, since the compiler knows all possible enumeration values.
* This allows us to access static information about the content type dynamically at runtime via a `Content` object. Previously this required a runtime check with branches for each content type, because static variables can't be resolved at runtime.
* The accessors in the `ContentMap` can now be type safe and don't have to be `null` checked.

Downsides:
* Unfortunately generic enums don't work right now. While we can declare them (e.g. ```enum ContentType<T extends Content> { song<Song>, ... }```, all `switch` statements will complain that they are not exhaustive, even though they are. This means we can't use the content type as a way to determine a generic parameter, instead we have to specify the generic argument explicitly (e.g. we cannot write `List<T> get<T extends Content>(ContentType<T> type)`, instead we have to specify the generic parameter when calling: `container.get<Song>(ContentType.song)`).
  Edit: It turns out we can use it in a couple places, as long as there is no `switch` on the variable, see 436f4e3fd2cab56c9ba43942ecb6cb9506abd56a.
  Edit 2: I just realized that we could also use it for the switch statements, if we cast to `ContentType`. This causes a `unnecessary_cast` warning, which we would need to supress:
   ```dart
   ContentType<T> contentType = ...;
   switch (contentType as ContentType) { // ignore: unnecessary_cast
   ```
   What do you think, should we do that as a workaround until dart is fixed? We could avoid explicitly specifying the generic type in a lot of places.
  Edit 3: There is already a bug report about this at the dart repo: dart-lang/language#2315

We don't loose any of the attributes that `contentPick` gave us by using `switch`es. When we calculate a result within the branches of the `switch`, it is a compile time error to miss one of the enum values. This covers most of the cases where `contentPick` was used. For the other cases, I enabled the `exhaustive_cases` warning and turned it's severity into an error. Unfortunately, this is not a build error, but it shows as en error for the linter. We could add a linter step to the CI, if you think that's worth doing.


lib/logic/player/content.dart
```diff
@@ -23,137 +22,71 @@ extension QuickActionSerialization on QuickAction {
...
-      Content.enumerate().contains(key ?? T),
-      "Specified type must be a subtype of Content",
+  /// The value for [ContentType.song].
+  V songValue;
+  /// The value for [ContentType.album].
+  V albumValue;
+  /// The value for [ContentType.playlist].
+  V playlistValue;
+  /// The value for [ContentType.artist].
+  V artistValue;
```
nt4f04uNd, line 34: This should be a `Map<ContenType, V>`. This would greatly reduce the implementation size.

lib/logic/player/content.dart
```diff
@@ -162,46 +95,40 @@ class ContentMap<V> {
...
-    _map.setValue<Album>(albums);
-    _map.setValue<Playlist>(playlists);
-    _map.setValue<Artist>(artists);
+  final List<Song> songs;
+  final List<Album> albums;
+  final List<Playlist> playlists;
+  final List<Artist> artists;
+
+  const ContentTuple(
+      {this.songs = const [], this.albums = const [], this.playlists = const [], this.artists = const []});
```
nt4f04uNd, line 104: trailing comma

lib/logic/player/content.dart
```diff
@@ -162,46 +95,40 @@ class ContentMap<V> {
...
-          return true;
-        }()) {
-    _map.setValue<Song>(songs);
-    _map.setValue<Album>(albums);
-    _map.setValue<Playlist>(playlists);
-    _map.setValue<Artist>(artists);
+  final List<Song> songs;
+  final List<Album> albums;
+  final List<Playlist> playlists;
+  final List<Artist> artists;
```
nt4f04uNd, line 101: I don't understand why this is no longer using a map.

lib/logic/player/content.dart
```diff
@@ -383,73 +310,75 @@ class ContentControl extends Control {
...
-      playlist: () => state.playlists as List<T>,
-      artist: () => state.artists as List<T>,
-    )();
+    return contentList;
   }
 
   /// Returns content of specified type with ID.
-  T? getContentById<T extends Content>(int id, [Type? contentType]) {
-    if ((contentType ?? T) == Album) {
+  T? getContentById<T extends Content>(int id, ContentType<T> contentType) {
```
nt4f04uNd, line 340: Oh, so it's able to imply the generics here. And it's also a compile time error to have something like `getContentById<Song>(entry.id, ContentType.album)`. Very cool.

nt4f04uNd:
> What do you think, should we do that as a workaround until dart is fixed? We could avoid explicitly specifying the generic type in a lot of places.

Seems like it's OK for us to suppress it, please just leave a TODO to clean it up with a link to the language issue in each such place

lib/logic/player/media_store_content_observer.dart
```diff
@@ -39,17 +35,18 @@ class MediaStoreContentObserver<T extends Content> extends ContentObserver {
...
 
-  String get _uri => contentPick<T, String>(
-        contentType: contentType,
-        // MediaStore.Audio.Media.EXTERNAL_CONTENT_URI
-        song: 'content://media/external/audio/media',
-        // MediaStore.Audio.Albums.EXTERNAL_CONTENT_URI
-        album: 'content://media/external/audio/albums',
-        // MediaStore.Audio.Playlists.EXTERNAL_CONTENT_URI
-        playlist: 'content://media/external/audio/playlists',
-        // MediaStore.Audio.Artists.EXTERNAL_CONTENT_URI
```
nt4f04uNd, line 50: Keep the comments

nt4f04uNd (changes requested):
Overall looking very good, although I didn't look closely at everything.

I found that going into the queue and selecting a song throws this

<details>

```
════════ Exception caught by widgets library ═══════════════════════════════════
type 'ContentSelectionController<SelectionEntry<Content>>' is not a subtype of type 'ContentSelectionController<SelectionEntry<Song>>' in type cast
The relevant error-causing widget was
RemoveFromQueueSelectionAction
lib\…\home_route\player_route.dart:44
════════════════════════════════════════════════════════════════════════════════
```
</details>

I have a feeling that this is definitely going to break something else, because the refactor is just too big, but it's for the best.

lib/logic/player/favorites.dart
```diff
@@ -19,20 +17,20 @@ class FavoritesControl with Control {
   @visibleForTesting
   final repository = FavoritesRepository();
 
-  final _favoriteSetsMap = ContentMap<Set<int>>();
+  final _favoriteSetsMap = ContentMap.fromFactory((_) => <int>{});
```
nt4f04uNd, line 20: Please change it to a meaningful parameter name, rather than `_`

lib/logic/player/content.dart
Abestanis: I would like to avoid using a `Map` here and in the `ContentTuple`, because

* A `Map` is (very slightly, but still) less efficient, accessing an item requires a runtime lookup. With the current way the compiler can resolve the item at compile time.
* Using a `Map` we would loose some of our type guarantees. We know that there is exactly one item for each content type, and the compiler should know too. With a map, we loose that information and have to force unwrap the value on access (e.g. `map[contentType]!`)
* At least for the `ContentTuple`, we still want the `songs`, `albums` ... accessors, so the code is actually not going to be much smaller.
* The explicit specification of a field for each content type makes sure we don'T forget to add a constructor argument when we add a new `ContentType`. Right now, the compiler would complain on the `switch` in `get` and `set`, which forces us to add a new field, which needs to be initialized so it forces us to add a new constructor argument. With a map, this chain would end at the `switch`.

Abestanis:
> Overall looking very good, although I didn't look closely at everything.

Thanks for the review, I know this one is not easy because it touches so many areas.

> I found that going into the queue and selecting a song throws this

Great catch! I added a golden test so this area is better covered. The problem is fixed by 43b81b7192f8d2b26f654f3232c829ae16ad3910, though I really don't know why the compiler is not able to interfere the correct generic type for the `ContentSelectionController` from the `contenTtype` argument. Maybe because it's nullable?

> I have a feeling that this is definitely going to break something else, because the refactor is just too big, but it's for the best.

I'm glad you see the value in this refactor, and also that we have tests that we have tests which would hopefully catch any major screw-up. I'm going to do a more throughout testing of all parts of the app that I can reach, so hopefully we'll catch any remaining problems.

nt4f04uNd:
> Maybe because it's nullable?

No. it's something else, I feel. I tested, and with `contentType: ContentType.song`, it creates `ContentSelectionController<SelectionEntry<Song>>`, but when we specify none, it creates `ContentSelectionController<SelectionEntry<Content>>` (which I was surprised it was able to pull off). So this needs a bit more of investigation.

Abestanis:
> with contentType: ContentType.song, it creates ContentSelectionController<SelectionEntry<Song>>,

But apparently not all the time. If you look at 43b81b7192f8d2b26f654f3232c829ae16ad3910, there is a `contentType: ContentType.song`, but the resulting `ContentSelectionController` evidently has `SelectionEntry<Content>` as it's generic parameter. Otherwise the cast wouldn't fail.

nt4f04uNd:
I suspect that you should make initSelectionController generic for this to be working

lib/routes/home_route/search_route.dart
```diff
@@ -117,31 +115,31 @@ class _SearchStateDelegate {
...
+        chipContextMap.get(value)!,
         duration: kTabScrollDuration,
         alignment: 0.5,
         alignmentPolicy: ScrollPositionAlignmentPolicy.explicit,
       );
     }
   }
 
-  ContentMap<BuildContext> chipContextMap = ContentMap();
+  ContentMap<BuildContext?> chipContextMap = ContentMap.fromFactory((_) => null);
```
nt4f04uNd, line 138: parameter name

lib/routes/home_route/search_route.dart
```diff
@@ -50,14 +50,12 @@ class ContentSearchDelegate {
 
 /// Search results container.
 class _Results {
-  ContentMap<List<Content>> map = ContentMap({
-    for (final contentType in Content.enumerate()) contentType: [],
-  });
+  ContentMap<List<Content>> map = ContentMap.fromFactory((_) => []);
```
nt4f04uNd, line 53: parameter name

lib/logic/player/favorites.dart
```diff
@@ -7,10 +7,8 @@ import 'package:sweyer/sweyer.dart';
 
 @visibleForTesting
 class FavoritesRepository {
-  final serializersMap = ContentMap<IntSerializerType>({
-    for (final contentType in Content.enumerate())
-      contentType: IntListSerializer('favorites_${ContentUtils.contentTypeId(contentType)}.json')
-  });
+  final serializersMap = ContentMap<IntSerializerType>.fromFactory(
+      (contentType) => IntListSerializer('favorites_${contentType.name}.json'));
```
nt4f04uNd, line 11: trailing comma

lib/logic/player/content.dart
```diff
@@ -23,137 +22,74 @@ extension QuickActionSerialization on QuickAction {
...
+        return albumValue;
+      case ContentType.playlist:
+        return playlistValue;
+      case ContentType.artist:
+        return artistValue;
+    }
   }
 
-  /// Removes all entries from the map.
-  void clear() {
```
nt4f04uNd, line 155: You could do

```dart
extension ContentMapClearExtension on ContentMap<Object?> {
  void clear() {
    songValue = null;
    albumValue = null;
    playlistValue = null;
    artistValue = null;
  }
}
```

lib/logic/player/favorites.dart
```diff
@@ -41,13 +39,15 @@ class FavoritesControl with Control {
 
   @override
   void dispose() {
-    _favoriteSetsMap.clear();
+    for (var element in _favoriteSetsMap.entries) {
+      element.value.clear();
+    }
```
nt4f04uNd, line 44: refactor this back, when you add ContentMapClearExtension

lib/logic/models/sort.dart
```diff
@@ -1,21 +1,25 @@
 import 'package:enum_to_string/enum_to_string.dart';
 import 'package:equatable/equatable.dart';
-import 'package:flutter/foundation.dart';
 import 'package:sweyer/sweyer.dart';
 
 /// Interface for other sort feature enums.
 abstract class SortFeature<T extends Content> extends Enum<String> {
```
nt4f04uNd, line 6: This could be refactored using enhanced enums as well, by the way. But let's keep this for another patch.

lib/widgets/selection.dart
```diff
@@ -1212,8 +1220,10 @@ class _SelectionCounterState extends State<SelectionCounter> with SelectionHandl
 //*********** Queue actions ***********
 
 /// Action that queues a [Song] or a [SongOrigin] to be played next.
-class _PlayNextSelectionAction<T extends Content> extends StatelessWidget {
-  const _PlayNextSelectionAction({Key? key}) : super(key: key);
+class _PlayNextSelectionAction extends StatelessWidget {
+  final ContentType? contentType;
+
+  const _PlayNextSelectionAction({this.contentType, Key? key}) : super(key: key);
```
nt4f04uNd, line 1226: I follow a style of that a constructors are at the top of the class

lib/widgets/selection.dart
```diff
@@ -1302,8 +1313,10 @@ class _PlayNextSelectionAction<T extends Content> extends StatelessWidget {
 }
 
 /// Action that adds a [Song] or a [SongOrigin] to the end of the queue.
-class _AddToQueueSelectionAction<T extends Content> extends StatelessWidget {
-  const _AddToQueueSelectionAction({Key? key}) : super(key: key);
+class _AddToQueueSelectionAction extends StatelessWidget {
+  final ContentType? contentType;
```
nt4f04uNd, line 1317: I follow a style of that a constructors are at the top of the class

nt4f04uNd:
Can you remove the updated goldens and just rebase to the master as well?

lib/logic/models/sort.dart
Abestanis: Yeah, one thing at a time. 😄

lib/widgets/selection.dart
Abestanis: There is a linter for this: [`sort_constructors_first`](https://dart-lang.github.io/linter/lints/sort_constructors_first.html). We should definitely enable it, but not in this PR, because a bunch of other classes don't follow the style already.

lib/logic/player/favorites.dart
Abestanis: Are you sure setting them to null is a good idea? This means I have to make the type nullable and nullcheck on access.

lib/logic/player/content.dart
Abestanis: I would do into
```dart
void clear() {
  for (final contentType in ContentType.values) {
    set(key: contentType, null);
  }
}
```
to avoid missing a ContentType when we add one.

https://github.com/nt4f04uNd/sweyer/pull/70#discussion_r912414199 would be the only usage of this. Let me know if you still want me to add this extension if you decide against setting them to `null`.

Abestanis: Also, the extension has a massive problem:

`Object?` is a parent type of everything, including non-null types.
<details>
<summary> Graphic from dart.dev </summary>

[Dart type hierarchy](https://dart.dev/null-safety/understanding-null-safety#top-and-bottom)
</details>

This means it matches everything, including non nullable types:
```dart
final ContentMap<int> x = ...;
final ContentMap<Object?> y = x;
x.clear();
```

This problem is not unique to `ContentMap`, I can do the same with a `List`:
```dart
final a List<int> = [1,2];
final List<Object?> b = a;
b.add(null);
```

nt4f04uNd: Yeah, you are right, I thought it would only work on nullable types.

lib/logic/player/favorites.dart
nt4f04uNd: Ok, let's keep it like this

lib/widgets/selection.dart
nt4f04uNd: The problem is that in some cases it actually is more valid and readable to have the constructors go last - e.g. see `Song` and other models.

test/golden/routes_golden_test.dart
```diff
@@ -198,6 +199,14 @@ void main() {
   });
 
   group('player_route', () {
+    /// Open the player queue screen.
+    Future<void> openPlayerQueue(WidgetTester tester) async {
```
nt4f04uNd, line 203: We already have similar functions in `player_route_test.dart`. Please place them into `test.dart` and re-use here.

lib/widgets/selection.dart
Abestanis: Yeah, I usually always put the member and fields first, I find it more readable. Let's not enable the lint rule then, you just going to have to tell me when I forget. 😃

test/golden/routes_golden_test.dart
Abestanis: 💯 Good idea. Though if we add any more of these helpers I would vote to put them in a separate file, like a `test_navigator.dart`, because `test.dart` is getting kind of crowded.

Abestanis:
Before you merge, let me just rename one of these helper functions for clarity.

test/golden/routes_golden_test.dart
nt4f04uNd: Yeah, sounds like a good idea

Abestanis:
Ups, looks like I switched up the comments during the copy, sorry, on sec

PR #71, opened 2022-06-25
Allow to keep favorites when enabling/disabling using the MediaStore favorites

Abestanis:
This improves the user experience when switching the `Use Android system for favorite tracks` setting. This is a continuation of my proposal in https://github.com/nt4f04uNd/sweyer/pull/34#discussion_r878749919.

<details>
<summary>With this change, when user enables the setting and there are conflicts between the apps favorite state and the MediaStore favorite state, a following dialog is displayed.</summary>

Conflict resolution dialog when enabling the MediaStore favorite backend
</details>

The dialog contains a list, where we display all songs which are locally marked as favorite, but not in the MediaStore with an enabled favorite indicator.
Below them, we show the songs which are marked as favorite in the MediaStore, but not locally with a disabled favorite indicator.
We don't show non-conflicting songs.
The user can then choose which of the songs should stay marked as favorite. When the user clicks the `Accept` button, we update the MediaStore if necessary. If the user clicks cancel or outside the dialog, we abort and disable the setting again.

When the user disables the setting, we save the currently marked favorites in the local database. No dialog is needed, because at this point the MediaStore is the only source of truth.


lib/localization/arb/intl_ru.arb
```diff
@@ -260,5 +260,7 @@
   "accept": "Принять",
   "cancel": "Отмена",
   "copied": "Скопировано",
-  "gitHubRepo": "GitHub репозиторий"
+  "gitHubRepo": "GitHub репозиторий",
+  "resolveConflict": "Разрешить конфликт",
+  "conflictExplanation": "В системе есть другие избранные файлы, отличные от Sweyer, выберите, какие из них следует оставить в качестве избранных. Показаны только конфликтующие песни."
```
Abestanis, line 265: Please have a look at these translations, those are only generated with GoogleTranslate.

nt4f04uNd: It lost its meaning a bit :-)
Let's make it "Избранные треки в системе отличаются от Sweyer, выберите, какие из них следует оставить в качестве избранных. Показаны только конфликтующие треки."

lib/logic/player/favorites.dart
```diff
@@ -65,16 +72,114 @@ class FavoritesControl with Control {
...
+      final context = AppRouter.instance.navigatorKey.currentContext;
+      if (context != null) {
+        final chosenFavorites =
+            await _showFavoriteConflictDialog(context, newFavoritesFromLocalStore, newFavoritesFromMediaStore);
+        if (chosenFavorites == null) {
+          await Settings.useMediaStoreForFavoriteSongs.set(false);
+          return false;
+        }
+        final songsToUnFavor = newFavoritesFromMediaStore.difference(chosenFavorites);
+        final songsToFavour = chosenFavorites.difference(newFavoritesFromMediaStore);
```
nt4f04uNd, line 103: let's be consistent with what dialect we use - we use American "favor" everwhere else 🙃
also, it should be "unfavor", no need to uppercase F

Here and everywhery else in this PR

test/logic/player/favorites_test.dart
```diff
@@ -0,0 +1,121 @@
...
+        Finder findInDialog(Finder finder) => find.descendant(of: find.byType(AlertDialog), matching: finder);
+        expect(findInDialog(find.text(l10n.resolveConflict)), findsOneWidget);
+        expect(findInDialog(find.text(notFavoriteInBothSong.title)), findsNothing);
+        expect(findInDialog(find.text(localFavoriteAndMediaStoreSong.title)), findsNothing);
+        expect(findInDialog(find.text(localFavoriteButNotInMediaStoreKeepSong.title)), findsOneWidget);
+        expect(findInDialog(find.text(localFavoriteButNotInMediaStoreUnFavorSong.title)), findsOneWidget);
+        expect(findInDialog(find.text(mediaStoreFavoriteButNotLocalKeepSong.title)), findsOneWidget);
+        expect(findInDialog(find.text(mediaStoreFavoriteButNotLocalUnFavorSong.title)), findsOneWidget);
+
+        // Un-favor the locally favored song
```
nt4f04uNd, line 64: no need for hyphen

nt4f04uNd (commented):
Hey @Abestanis, this is an awesome PR, thanks for making this!
I'm sorry it took a long time to get to this, I will try to have more pace on this project further on.

I noted down a few grammatical things to change, could you also rebase this and I'll merge it?

Also did you test it in the wild manually? I can't seem to find any other app that does this sort of thing to simulate this situation, but I recall that I definitely tested it somehow, when I added MediaStore source. Maybe I was using android_content_provider to do this...

nt4f04uNd:
> Also did you test it in the wild manually? I can't seem to find any other app that does this sort of thing to simulate this situation, but I recall that I definitely tested it somehow, when I added MediaStore source. Maybe I was using android_content_provider to do this...

Lol just after I posted this, I immediately figured it. I didn't fully simulate this, I just enabled `notifyForDescendants` option and that was enough to test this. Also to test your PR I now see it is enough to just mark a few songs as favorite and then enable MediaStore source.

Abestanis:
Hey @nt4f04uNd, nice to hear from you again.

> I noted down a few grammatical things to change, could you also rebase this and I'll merge it?

Thanks for the review, I rebased and implemented your remarks.

> it is enough to just mark a few songs as favorite and then enable MediaStore source.

Yeah, you can test all of the situations by marking/unmarking the songs as favorites in Sweyer and then switching between the backends. There is no difference if you marked a song as favorite locally and then switch to the MediaStore, or if some other app unmarks the song as favorite while you are using the local storage and you switch back afterwards.

But I did test this with a file manager (I think it was ` Files by Google`) on my actual device. In the filemanager, you can mark files as favorites and it updates the MediaStore.

nt4f04uNd:
> But I did test this with a file manager (I think it was ` Files by Google`) on my actual device. In the filemanager, you can mark files as favorites and it updates the MediaStore.

Interestengly enough, this doesn't work on my device for some reason

PR #84, opened 2023-04-07
Request the READ_MEDIA_AUDIO permission

### LINKED ISSUE #83: Sweyer no longer has access to music
### Steps to reproduce

1. Open the newest 1.0.8 version on a phone with Android 13
2. The app is not asking to grant storage permissions and the app shows that there is no music on my phone.
3. Deleted the app data and cache and opened it again.
4. This time the app askes to grant storage permissions.
5. The user is redirected to the settings page to grant the permission manually, but there is no option to grant any permission. The settings app shows that the app does not require any permission.

### What did you expect?

To see music and to be able to grant the app access to it.

### What happened instead?

The app does not have access to the music.

### Sweyer version

1.0.8

### Android version

Android 13

Abestanis:
On Android 13, `READ_EXTERNAL_STORAGE` has been split into more [granular permissions](https://developer.android.com/about/versions/13/behavior-changes-13#granular-media-permissions), like `READ_MEDIA_AUDIO`. We need to request this permission to be able to query music from the content provider service. Fixes #83.

The `READ_MEDIA_AUDIO` will automatically be granted when requested if `READ_EXTERNAL_STORAGE` was previously granted.

### DIFF
--- a/lib/logic/permissions.dart
+++ b/lib/logic/permissions.dart
@@ -5,30 +5,29 @@ import 'package:permission_handler/permission_handler.dart';
 class Permissions {
   static Permissions instance = Permissions();
 
-  /// Whether storage permission is granted
-  late PermissionStatus _permissionStorageStatus;
+  /// Whether audio access permission is granted
   late PermissionStatus _permissionAudioStatus;
 
   /// Returns true if permissions were granted
-  bool get granted => _permissionStorageStatus == PermissionStatus.granted
-      || _permissionAudioStatus == PermissionStatus.granted;
+  bool get granted => _permissionAudioStatus == PermissionStatus.granted;
 
   /// Returns true if permissions were not granted
   bool get notGranted => !granted;
 
+  /// The permission to use when requesting access to the audio files on the device
+  Permission get _audioPermission =>
+      DeviceInfoControl.instance.useAudioPermission ? Permission.audio : Permission.storage;
+
   Future<void> init() async {
-    _permissionStorageStatus = await Permission.storage.status;
-    _permissionAudioStatus = await Permission.audio.status;
+    _permissionAudioStatus = await _audioPermission.status;
   }
 
   Future<void> requestClick() async {
     final responses = await [Permission.storage, Permission.audio].request();
-    _permissionStorageStatus = responses[Permission.storage] ?? PermissionStatus.denied;
-    _permissionAudioStatus = responses[Permission.audio] ?? PermissionStatus.denied;
+    _permissionAudioStatus = responses[_audioPermission] ?? PermissionStatus.denied;
     if (granted) {
       await ContentControl.instance.init();
-    } else if (_permissionStorageStatus == PermissionStatus.permanentlyDenied ||
-        _permissionAudioStatus == PermissionStatus.permanentlyDenied) {
+    } else if (_permissionAudioStatus == PermissionStatus.permanentlyDenied) {
       final l10n = staticl10n;
       await ShowFunctions.instance.showToast(
         msg: l10n.allowAccessToExternalStorageManually,
--- a/lib/logic/permissions.dart
+++ b/lib/logic/permissions.dart
@@ -7,22 +7,28 @@ class Permissions {
 
   /// Whether storage permission is granted
   late PermissionStatus _permissionStorageStatus;
+  late PermissionStatus _permissionAudioStatus;
 
   /// Returns true if permissions were granted
-  bool get granted => _permissionStorageStatus == PermissionStatus.granted;
+  bool get granted => _permissionStorageStatus == PermissionStatus.granted
+      || _permissionAudioStatus == PermissionStatus.granted;
 
   /// Returns true if permissions were not granted
   bool get notGranted => !granted;
 
   Future<void> init() async {
     _permissionStorageStatus = await Permission.storage.status;
+    _permissionAudioStatus = await Permission.audio.status;
   }
 
   Future<void> requestClick() async {
-    _permissionStorageStatus = await Permission.storage.request();
-    if (_permissionStorageStatus == PermissionStatus.granted) {
+    final responses = await [Permission.storage, Permission.audio].request();
+    _permissionStorageStatus = responses[Permission.storage] ?? PermissionStatus.denied;
+    _permissionAudioStatus = responses[Permission.audio] ?? PermissionStatus.denied;
+    if (granted) {
       await ContentControl.instance.init();
-    } else if (_permissionStorageStatus == PermissionStatus.permanentlyDenied) {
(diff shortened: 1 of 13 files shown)


Abestanis:
Making the tests pass again was harder than the actual fix. 😄

This fix was tested on an Android 13 phone and an emulator with Android 10.
I think we should make a `1.8.1` release as soon as possible, to make the app usable on Android 13+ phones again.

test/routes/home_route_test.dart
```diff
@@ -13,53 +13,75 @@ void main() {
...
-  testWidgets('permissions screen - shows toast and opens settings when permissions are denied',
-      (WidgetTester tester) async {
-    late PermissionsChannelObserver permissionsObserver;
-    await setUpAppTest(() {
-      permissionsObserver = PermissionsChannelObserver(tester.binding);
-      permissionsObserver.setPermission(Permission.storage, PermissionStatus.denied);
+    // On Android 13+ the storage permission is removed and reports as permanentlyDenied.
+    // If the user granted audio permissions, we have access to the music
+    // and don't want to show the permission request screen.
+    testWidgets('does not show when one removed permission is permanently denied', (WidgetTester tester) async {
```
nt4f04uNd, line 46: why not test it for both?
on API `<33` he `Permission.audio` is permantently denied

test/fakes/fake_just_audio.dart
```diff
@@ -137,8 +142,10 @@ class MockAudioPlayer implements AudioPlayerPlatform {
     }
     _audioSource = audioSource;
     _index = request.initialIndex ?? 0;
-    // Simulate loading time.
-    await Future<dynamic>.delayed(const Duration(milliseconds: 100));
+    if (binding.inTest) {
+      // Simulate loading time.
+      await Future<void>.delayed(const Duration(milliseconds: 100));
```
nt4f04uNd, line 147: Can you explain this change?

lib/logic/permissions.dart
```diff
@@ -7,22 +7,28 @@ class Permissions {
...
 
   Future<void> init() async {
     _permissionStorageStatus = await Permission.storage.status;
+    _permissionAudioStatus = await Permission.audio.status;
   }
 
   Future<void> requestClick() async {
-    _permissionStorageStatus = await Permission.storage.request();
-    if (_permissionStorageStatus == PermissionStatus.granted) {
+    final responses = await [Permission.storage, Permission.audio].request();
```
nt4f04uNd, line 25: lol, cool 👍

lib/logic/permissions.dart
```diff
@@ -7,22 +7,28 @@ class Permissions {
...
-  bool get granted => _permissionStorageStatus == PermissionStatus.granted;
+  bool get granted => _permissionStorageStatus == PermissionStatus.granted
+      || _permissionAudioStatus == PermissionStatus.granted;
 
   /// Returns true if permissions were not granted
   bool get notGranted => !granted;
 
   Future<void> init() async {
     _permissionStorageStatus = await Permission.storage.status;
+    _permissionAudioStatus = await Permission.audio.status;
```
nt4f04uNd, line 21: maybe we should use DeviceInfo API level call storage status only `<33`, and audio status only `>=33`?

test/test.dart
```diff
@@ -189,7 +189,11 @@ Future<void> setUpAppTest([VoidCallback? configureFakes]) async {
   ThemeControl.instance.init();
   ThemeControl.instance.initSystemUi();
   await Permissions.instance.init();
-  await ContentControl.instance.init();
+  if (binding.inTest) {
+    await binding.runAsync(() => ContentControl.instance.init());
```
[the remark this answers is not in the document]
Abestanis: Because we are doing real async work with the `await Future<void>.delayed(const Duration(milliseconds: 100));` in `test/fakes/fake_just_audio.dart:147`

This caused the test to hang (see [here](https://github.com/nt4f04uNd/sweyer/actions/runs/4644452852/jobs/8219736699#step:3:875)).

Originally, I wanted to is `binding.delayed` instead of `Future.delayed`, but because this is apparently executed in the background while the test is running, it threw an error about a reentrant use of the delayed function.

test/fakes/fake_just_audio.dart
Abestanis: See, my other response. I added the `binding.inTest` check because it occurred to me that outside of the test we don't need to delay.

test/routes/home_route_test.dart
```diff
@@ -13,53 +13,75 @@ void main() {
...
+    testWidgets('shows toast and opens settings when permissions are denied', (WidgetTester tester) async {
+      late PermissionsChannelObserver permissionsObserver;
+      await setUpAppTest(() {
+        permissionsObserver = PermissionsChannelObserver(tester.binding);
+        permissionsObserver.setPermission(Permission.storage, PermissionStatus.denied);
+        permissionsObserver.setPermission(Permission.audio, PermissionStatus.denied);
+      });
+      await tester.runAppTest(() async {
+        permissionsObserver.setPermission(Permission.storage, PermissionStatus.permanentlyDenied);
+        permissionsObserver.setPermission(Permission.audio, PermissionStatus.permanentlyDenied);
```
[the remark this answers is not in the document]
Abestanis: It was like this before, but we are testing the behaviour when the user denies the system dialog.

lib/logic/permissions.dart
Abestanis: We want to request both of them at the same time to automatically get the `audio` permission if `storage` was previously granted.

test/test.dart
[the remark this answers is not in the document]
nt4f04uNd: It appears to me it timed out there, because you didn't update this specific permission test. But I think I'm okay with this change overall.

test/routes/home_route_test.dart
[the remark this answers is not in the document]
nt4f04uNd: Seems like it could just be one call in set up step, but skipping this in this PR

lib/logic/permissions.dart
nt4f04uNd: Didn't really get what you mean, but so be it.

nt4f04uNd:
I'll be able to re publish the app in a few hours. Waiting for additional test on API `<33` and will merge this.

nt4f04uNd:
Also, please iterate the version in pubspec and update the CHANGELOG

lib/logic/permissions.dart
Abestanis: Ah, sorry, I was in a hurry and didn't really read what you were writing. You are right, it is better to only check for the required permission, I changed it. What I meant is that we can (and I think we should) request both permissions simultaneously.

PR #78, opened 2022-07-14
Move the native Android code of Sweyer into a plugin

Abestanis:
*I'm opening this as a draft so you can see what I'm working on right now. Once the remaining issues are resolved I will mark this as ready for review.*

This is a rather big change, sorry about that. This is work towards the homescreen widget.

We move all of the Android specific Java code into a plugin called `sweyer_plugin`.
During the migration, I converted the former `ContentChannel.java`  to Kotlin and merged it into `SweyerPlugin.kt`, because it was easier that way.

This fixes #73. The problem was that if Sweyer was not started via the main activity, then the `content_channel` `MethodChannel` wouldn't be registered, causing the app to throw on startup.

This can be replicated as follows:
1. Make sure Sweyer is not started and that there is no cached `FlutterEngine`. This is the case after a long time of not using the app, a phone reboot or after force stopping Sweyer.
2. Emit a media play/pause key press. This can be done by attaching a physical device with a play/pause button to the phone, like a headset, and pressing the button. Alternatively, we can use the [Android Debug Bridge](https://developer.android.com/studio/command-line/adb) to emit the corresponding keycode ([`KEYCODE_MEDIA_PLAY_PAUSE`, `85`](https://developer.android.com/reference/android/view/KeyEvent#KEYCODE_MEDIA_PLAY_PAUSE)):
    ```shell
    adb shell input keyevent 85
    ```
3. Observe that no music is played and the following error can be seen in Logcat:
    <details>
    <summary>Stacktrace</summary>

    ```dart
    MissingPluginException(No implementation found for method retrieveSongs on channel content_channel)
    #0      MethodChannel._invokeMethod (package:flutter/src/services/platform_channel.dart:165:7)
    <asynchronous suspension>
    #1      MethodChannel.invokeListMethod (package:flutter/src/services/platform_channel.dart:353:35)
    <asynchronous suspension>
    #2      ContentChannel.retrieveSongs (package:sweyer/logic/player/content_channel.dart:113:18)
    <asynchronous suspension>
    #3      ContentControl.refetch (package:sweyer/logic/player/content.dart:383:33)
    <asynchronous suspension>
    #4      Future.wait.<anonymous closure> (dart:async/future.dart:521:21)
    <asynchronous suspension>
    #5      Future.any.onValue (dart:async/future.dart:611:5)
    <asynchronous suspension>
    ```
    </details>
4. In addition to this, the app can also no longer be started normally (via the launcher), it get's stuck at the splash screen.

[...]

Issue #81, opened 2022-12-02
Use riverpod


nt4f04uNd:
The current situation with dependency injection is rather horrible.
App should be rewritten with DI in mind.

Currently app mostly uses singletones as a main architecture solution, which was a bad decision in the hindsight. There aren't a lot of business logic classes, so the problem might not be immediately obvious:

* It obfuscates the dependencies of concrete classes, which in its turn makes it very hard to write isolated tests, mocks, fakes and generally judge what a class does.
* This also allows circular dependencies, which also might make the repsonsibility of a class a bit fuzzy, and leads to convoluted and unbovious business logic, order of execution, etc.

A good option for a DI, which I personally like is [riverpod](https://riverpod.dev/).
It is easy, flexible, scalable, easily mockable and doesn't force you into doing some sketchy things, like for example Provider does by forcing you to tie your dependencies to a widget tree.

nt4f04uNd:
There are even global variables, which of course should be removed. Having them allows for state-leakages between tests.
I left a TODO in code linking to this issue, but for now it is OK to leave them there.

nt4f04uNd:
Started doing this in https://github.com/nt4f04uNd/sweyer/pull/79

nt4f04uNd:
We should use https://github.com/yandex/yx_scope instead

Abestanis:
The current approach with singletons also makes testing correctly difficult. We have futures that we are not awaiting, so they can continue to execute after the test is finished and into the next test, where they can then interact with the fakes of the new test accidentally.   A proper dependency injection mechanism should provide a way to scope the dependencies, ideally to a widget tree, so we can avoid this problem.

PR #85, opened 2023-04-25
Move the native Android code of Sweyer into a plugin

### LINKED ISSUE #73: Reacting to media play button doesn't work when Sweyer is not running
### Steps to reproduce

1. Make sure Sweyer is not running (force close the app).
2. Trigger a [`KEYCODE_MEDIA_PLAY` event](https://developer.android.com/reference/android/view/KeyEvent#KEYCODE_MEDIA_PLAY) or [`KEYCODE_MEDIA_PLAY_PAUSE` event](https://developer.android.com/reference/android/view/KeyEvent#KEYCODE_MEDIA_PLAY_PAUSE), e.g. press the play button on a connected headset.
3. Observe the logcat logs.

### What did you expect?

The app should be able to start from the `audio_service` and start playing audio.

### What happened instead?

The app does start in the background, but never plays audio. This is because the following error is thrown:
<details>
<summary>Error</summary>

```dart
I/System.out: ### onAttachedToEngine
I/System.out: ### 1 client handlers
I/System.out: ### new AudioHandlerInterface
I/System.out: ### connect
D/MediaBrowserCompat: Connecting to a MediaBrowserService.
I/System.out: ### connect returned
I/System.out: ### onAttachedToEngine completed
D/DeviceInfoPlugin: Use TaskQueues.
I/System.out: flutterEngine warmed up
I/System.out: ### onStartCommand
I/System.out: ### onMediaButtonEvent: KeyEvent { action=ACTION_DOWN, keyCode=KEYCODE_MEDIA_PLAY_PAUSE, scanCode=0, metaState=0, flags=0x0, repeatCount=0, eventTime=0, downTime=0, deviceId=-1, source=0x0, displayId=0 }
2022-06-26 17:39:10.733 6154-6154/com.nt4f04und.sweyer I/System.out: ### listener = com.ryanheise.audioservice.AudioServicePlugin$AudioHandlerInterface@8194acc
I/System.out: ### calling onClick
I/System.out: ### sending click map: {button=0}
I/System.out: ### called onClick
I/System.out: ### onGetRoot. isRecentRequest=false
I/System.out: ### onConnected
I/System.out: ### registered mediaController callback
I/System.out: ### onConnected returned
I/flutter: The Dart VM service is listening on http://127.0.0.1:45025/VEcJ_HFwVvI=/
I/flutter: ----------------FIREBASE CRASHLYTICS----------------
I/flutter: MissingPluginException(No implementation found for method retrieveSongs on channel content_channel)
I/flutter: #0      MethodChannel._invokeMethod (package:flutter/src/services/platform_channel.dart:165:7)
    <asynchronous suspension>
    #1      MethodChannel.invokeListMethod (package:flutter/src/services/platform_channel.dart:353:35)
    <asynchronous suspension>
    #2      ContentChannel.retrieveSongs (package:sweyer/logic/player/content_channel.dart:113:18)
    <asynchronous suspension>
    #3      ContentControl.refetch.<anonymous closure> (package:sweyer/logic/player/content.dart:444:33)
    <asynchronous suspension>
    #4      ContentControl.refetch (package:sweyer/logic/player/content.dart:441:5)
    <asynchronous suspension>
    #5      Future.wait.<anonymous closure> (dart:async/future.dart:521:21)
    <asynchronous suspension>
    #6      Future.any.onValue (dart:async/future.dart:611:5)
    <asynchronous suspension>
```
</details>

The reason for the error is that `ContentChannel` initialization is triggered by `MainActivity.onCreate`, but when the app is not running and we are starting via the audio service, no `MainActivity` is ever created and therefore the `MethodChannel` is never registered.

The only workaround I found (suggested by https://github.com/ryanheise/audio_service/issues/399#issuecomment-657326724) is to extract the method channel into a separate plugin and use the plugin registration system to initialize the method channel under all conditions.

@nt4f04uNd do you have a better idea how to approach this?

### Sweyer version

Latest on the `1.0.8` branch; c6386e62fcae56165233ab1c09e0d7f024075cc2

### Android version

_No response_

Abestanis:
*This is a continuation of #78, which I can't reopen.*

Fix #73 by moving all of the Android specific Java code into a plugin called `sweyer_plugin`. The former `ContentChannel.java` is merged into `SweyerPlugin.java`.

When Sweyer is not started via the main activity, the `content_channel` `MethodChannel` is not registered. This can happen when Sweyer is started via the `AudioService` by the `audio_service` package. Using a plugin works because plugins are automatically initialized via the `GeneratedPluginRegistrant` when a `FlutterEngine` is initialized, no matter how that engine is created.

The problem can be replicated as follows:
* Force stop Sweyer to make sure that there is no cached FlutterEngine.
* Emit a media play/pause key press. This can be done by attaching a physical device with a play/pause button to the phone, like a headset, and pressing the button. Alternatively, we can use the [Android Debug Bridge](https://developer.android.com/studio/command-line/adb) to emit the corresponding keycode ([KEYCODE_MEDIA_PLAY_PAUSE, 85](https://developer.android.com/reference/android/view/KeyEvent#KEYCODE_MEDIA_PLAY_PAUSE)):
  ```shell
  adb shell input keyevent 85
  ```
* Observe:
  - Without these changes: No music is played and there is an error in Logcat.
  - With these changes: Sweyer starts playing music and there is no error in Logcat.
* When starting the app afterwards:
  - Without these changes: Sweyer is stuck at the splash screen and requires a force stop to start again.
  - With these changes: Sweyer starts normally and correctly displays the currently playing song.

<details>
<summary>Video of the steps with and without the changes</summary>

| Without the changes | With the changes |
| ----------------------- | ------------------- |
| Reproduction steps without the fix | Reproduction steps with the fix |

</details>


lib/logic/models/playlist.dart
```diff
@@ -1,14 +1,15 @@
 import 'package:audio_service/audio_service.dart';
 import 'package:sweyer/sweyer.dart';
+import 'package:sweyer_plugin/sweyer_plugin.dart';
 
-class Playlist extends PersistentQueue with DuplicatingSongOriginMixin {
+class Playlist extends PersistentQueue with DuplicatingSongOriginMixin, MediaStorePlaylist {
```
Abestanis, line 5: > we shouldn't mix-in classes

What's the reason for not using mix-in classes?

Abestanis:
> It unnecessarily converts Java's code to Kotlin's while also moving it to a different location, which makes it very hard to review the actual changes. If you want to convert Java to Kotlin (which is also an arguable investment of my and your time but OK), please make it in a separate pull request.

I have reverted the conversion to Kotlin and will put it into a separate pull request.

> For some reason it uses a federated plugin structure. I don't see how this is justified, since the only reason it exists is to enable plugin athors to delegate implementation of a certain platforms for plugins to 3d parties, as well as allowing multiple implementations of those platforms. In our case it doesn't bring any benefit, while enables a lot of unncessary boilerplate.

We need to use a plugin, because that is the core of the fix. I took the official template of a flutter plugin, would you have wanted me to use a different template? In any case, I think this structure will make it easier to port Sweyer to Windows. The plugin provides us a clear interface of all the platform specific Android code and it gives us a way of providing the same interface for Windows. In my opinion this is worth the boilerplate, and probably even necessary.

android/app/src/main/java/com/nt4f04und/sweyer/MainActivity.java
```diff
@@ -1,50 +1,17 @@
...
 
-import com.nt4f04und.sweyer.channels.ContentChannel;
-import com.nt4f04und.sweyer.handlers.GeneralHandler;
 import com.ryanheise.audioservice.AudioServicePlugin;
 
 import io.flutter.embedding.android.FlutterActivity;
 import io.flutter.embedding.engine.FlutterEngine;
-import io.flutter.plugin.common.BinaryMessenger;
 
 public class MainActivity extends FlutterActivity {
```
Abestanis, line 12: At this point we could just extend from [`AudioServiceActivity`](https://github.com/ryanheise/audio_service/blob/minor/audio_service/android/src/main/java/com/ryanheise/audioservice/AudioServiceActivity.java).

lib/media_query_wrapper.dart
```diff
@@ -1,26 +0,0 @@
...
-  });
-
-  final Widget child;
-
-  @override
-  Widget build(BuildContext context, WidgetRef ref) {
-    final textScaleFactor = ref.watch(textScaleFactorStateNotifierProvider.select((value) => value));
-
-    return MediaQuery(
-      data: MediaQueryData.fromWindow(WidgetsBinding.instance.window).copyWith(
```
Abestanis, line 20: The `MediaQueryWrapper` was not updating the `MediaQueryData` correctly.

In the function documentation of `MediaQueryData.fromWindow` it says:
```
If you use this, you should ensure that you also register for
notifications so that you can update your [MediaQueryData] when the
window's metrics change. For example, see
[WidgetsBindingObserver.didChangeMetrics] or
[dart:ui.PlatformDispatcher.onMetricsChanged].
```
This was fixed by using the MediaQuery provided by `MaterialApp.router`, which updated correctly.

lib/routes/routes.dart
```diff
@@ -365,63 +366,74 @@ class AppRouter extends RouterDelegate<AppRoutes<Object?>>
...
-              transitionSettings: transitionSettings.greyDismissible,
-            ));
-          }
+    return Consumer(
+      builder: (context, ref, child) {
+        final mediaQuery = MediaQuery.of(context);
+        final textScaleFactor = ref.watch(textScaleFactorStateNotifierProvider.select((value) => value));
+        if (mediaQuery.size.isEmpty) {
+          return const SizedBox.shrink(); // Don't render the app if we are started in the background.
         }
```
Abestanis, line 375: When started in the background via a media play button, the `AudioService` will start the app without a window and `mediaQuery.size` will be empty. If we don't do an early return here we would do a lot of unnecessary work and also throw some divide by zero errors because of the `(0, 0)` window size.

When the app is started in the foreground the `MediaQuery` is updated and the app is drawn as normal.

test/golden/goldens/selection_route.selection_route_settings.light.png
Abestanis, line 1: It looks like this fixes a regression: The config icon is not supposed to be visible since #50 was merged.

I'm not exactly sure why it doesn't show up in the golden image. I can verify that the icon is visible in `1.0.9` and also visible with the changes in this pull request. I will investigate in another pull request.

test/test.dart
```diff
@@ -39,12 +40,10 @@ final _testSong = Song(
   dateModified: 0,
   duration: 0,
   size: 0,
-  data: 'data_data_data_data_data_data_data_data',
+  filesystemPath: '/path/to/song',
   isFavoriteInMediaStore: false,
   generationAdded: 0,
   generationModified: 0,
-  origin: _testAlbum,
-  duplicationIndex: 0,
```
Abestanis, line 47: This is the reason from most of the golden image changes.

The `_testSong` is used as a base for the songs returned from the Sweyer plugin (former content channel), but the `origin` and `duplicationIndex` of items returned by those will always be `null`, because those are an app level concept that is only added once the songs are added to a queue.

sweyer_plugin/lib/sweyer_plugin.dart
```diff
@@ -0,0 +1,183 @@
...
+/// Class to cancel the [ContentChannel.loadAlbumArt].
+class CancellationSignal {
+  CancellationSignal() : _id = _uuid.v4();
+  final String _id;
+
+  /// Cancel loading of an album art.
+  Future<void> cancel() => SweyerPluginPlatform.instance.cancelAlbumArtLoad(id: _id);
+}
+
+abstract class SweyerPlugin {
```
nt4f04uNd, line 71: Let's not make it abstract with static methods, but rather a regular class and expose it as singletone for now (until we have a proper DI setup in the app).

As it was done before.

sweyer_plugin/lib/sweyer_plugin.dart
```diff
@@ -0,0 +1,183 @@
...
+  /// See https://stackoverflow.com/questions/18606007/android-image-files-deleted-from-com-android-providers-media-albumthumbs-on-rebo
+  static Future<void> fixAlbumArt(int albumId) async => await SweyerPluginPlatform.instance.fixAlbumArt(albumId);
+
+  static Future<Iterable<T>> retrieveSongs<T extends MediaStoreSong>(
+    T Function(Map<String, dynamic> data) factory,
+  ) async =>
+      (await SweyerPluginPlatform.instance.retrieveSongs()).map(factory);
+
+  static Future<Map<int, T>> retrieveAlbums<T extends MediaStoreAlbum>(
+    T Function(Map<String, dynamic> data) factory,
```
nt4f04uNd, line 97: Let's factor this function signature into a `typedef _DataFactory<T> = T Function(Map<String, dynamic> data);`

sweyer_plugin/lib/sweyer_plugin.dart
```diff
@@ -0,0 +1,183 @@
...
+import 'dart:ui';
+
+import 'package:equatable/equatable.dart';
+import 'package:sweyer_plugin/sweyer_plugin_method_channel.dart';
+import 'package:uuid/uuid.dart';
+
+import 'sweyer_plugin_platform_interface.dart';
+
+/// A song from the native media store.
+abstract class MediaStoreSong {
```
nt4f04uNd, line 11: If we are abstracting out from the MediaStore, wouldn't it make more sense to call this something like RawSong or AbstractSong?

Here and for other classes

sweyer_plugin/lib/sweyer_plugin.dart
```diff
@@ -0,0 +1,183 @@
...
+  ///  * [SweyerPluginChannelException.playlistNotExists] when the playlist doesn't exist.
+  static Future<void> insertSongsInPlaylist({
+    required int index,
+    required List<MediaStoreSong> songs,
+    required MediaStorePlaylist playlist,
+  }) {
+    assert(songs.isNotEmpty);
+    assert(index >= 0 && index <= playlist.songIds.length + 1);
+    return SweyerPluginPlatform.instance.insertSongsInPlaylist(
+        index: index, songIds: songs.map((song) => song.sourceId).toList(), playlistId: playlist.id);
```
nt4f04uNd, line 161: trailing comma would make sense here

sweyer_plugin/lib/sweyer_plugin.dart
```diff
@@ -0,0 +1,183 @@
...
+  int get id;
+
+  /// A list of ids of songs that this playlist contains.
+  List<int> get songIds;
+}
+
+/// A genre from the native media store.
+abstract class MediaStoreGenre {}
+
+abstract class SweyerPluginException with EquatableMixin {
```
nt4f04uNd, line 37: All the abstract interfaces should go into the sweyer_plugin_platform_interface.dart file, which represents a core platform independent abstraction.

To better understand why let me refer to the documentation https://docs.flutter.dev/packages-and-plugins/developing-packages#federated-plugins

In our case:
- `sweyer_plugin.dart` is like the **app-facing package**
- `android` folder is like the **platform package**
- `sweyer_plugin_method_channel.dart` and `sweyer_plugin_platform_interface.dart` is the **platform interface**

So, with the current location of this class, we mix these layers in between, because now the **platform interface** (SweyerMethodChannelException) extends from something that is located in the app facing package

test/test.dart
nt4f04uNd: Yes thanks, it makes it more in line with a real environment.

test/golden/goldens/selection_route.selection_route_settings.light.png
nt4f04uNd: Given the https://github.com/nt4f04uNd/sweyer/pull/90, I think this test is just flaky without a StatefulBuilder. There are a lot of animations happening during opening this route and it could previously work and then stop working for whatever reason.

nt4f04uNd (commented):
> This is a continuation of #78, which I can't reopen.

It was closed automatically upon merging the target branch.

> We need to use a plugin, because that is the core of the fix. I took the official template of a flutter plugin, would you have wanted me to use a different template? In any case, I think this structure will make it easier to port Sweyer to Windows. The plugin provides us a clear interface of all the platform specific Android code and it gives us a way of providing the same interface for Windows. In my opinion this is worth the boilerplate, and probably even necessary.

I didn't say we don't need a plugin.

There are two plugin structures, generally:
- regular - all the code is just one MethodChannel and the native code to it
- federated - https://docs.flutter.dev/packages-and-plugins/developing-packages#federated-plugins

Federated plugins consist of multiple packages that fall into multiple categories (app-facing package, platform package(s), platform interface package). For example, this enables [non-endorsed plugin implementations (not authored by the plugin author)](https://docs.flutter.dev/packages-and-plugins/developing-packages#non-endorsed-federated-plugin) to exists.

My bad - from glancing at the code and seeing a platform interface, without actually reading through it, I thought that you created multiple packages and used the federated plugin structure. This in our case would just create for us more packages to manage, since we don't plan to publish it or reuse anywhere else.

I also didn't know that platform interface is now a default plugin template.

Overall, I now agree that what you did here is right. The interface and abstraction is good.

sweyer_plugin/lib/sweyer_plugin.dart
Abestanis: Yeah, let's rename it. I would prefer `PlatformSong`, because this is the representation of the song coming from the platform we are running on. Is that ok?

nt4f04uNd: Yeah

lib/routes/routes.dart
```diff
@@ -365,63 +366,74 @@ class AppRouter extends RouterDelegate<AppRoutes<Object?>>
...
-              ),
-              transitionSettings: transitionSettings.greyDismissible,
-            ));
-          }
+    return Consumer(
+      builder: (context, ref, child) {
+        final mediaQuery = MediaQuery.of(context);
+        final textScaleFactor = ref.watch(textScaleFactorStateNotifierProvider.select((value) => value));
+        if (mediaQuery.size.isEmpty) {
+          return const SizedBox.shrink(); // Don't render the app if we are started in the background.
```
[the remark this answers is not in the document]
Abestanis: I think we need to, yes. `main` is called when the `audio_service` initializes the FlutterEngine while we are in the background, and it is not called a second time when we go in the foreground. We need a way to initialize our UI when we switch from background to foreground, and I have been using the dependency on the `MediaQuery` here together with the fact that the screen size updates when we switch to the foreground to do that.

There is also the initialization of `NFWidgets` as well as initialization of the whole language and translation framework via the `supportedLocales` of `MaterialApp.router`. I'm not entirely sure if we need that while we are in the background, but the more we don't initialize as normal the higher the risk is that we try to use something uninitialized in one of the callbacks we get from the audio service side.

sweyer_plugin/example/test/widget_test.dart
```diff
@@ -0,0 +1,26 @@
...
+void main() {
+  testWidgets('Verify Platform version', (WidgetTester tester) async {
+    // Build our app and trigger a frame.
+    await tester.pumpWidget(const MyApp());
+
+    // Verify that platform version is retrieved.
+    expect(
+      find.byWidgetPredicate(
+        (Widget widget) => widget is Text && widget.data!.startsWith('Running on:'),
+      ),
```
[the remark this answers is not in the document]
Abestanis: Let's keep it. My hope is that this will be helpful when writing the Windows implementation.

lib/routes/routes.dart
```diff
@@ -365,63 +366,74 @@ class AppRouter extends RouterDelegate<AppRoutes<Object?>>
...
-            pages.add(StackFadePage(
-              key: AppRoutes.selection.uniqueKey,
-              child: SelectionRoute(
-                selectionArguments: (route as AppRoutes<SelectionArguments>).arguments!,
-              ),
-              transitionSettings: transitionSettings.greyDismissible,
-            ));
-          }
+    return Consumer(
+      builder: (context, ref, child) {
```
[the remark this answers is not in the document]
Abestanis: Because the `MediaQuery`, which this `Consumer` needs to use, is no longer available outside of the `MaterialApp.router`.
See the commit message of 57efb33d1280e9fef3947a69ab9f0209ee3148dd for context.

lib/logic/models/playlist.dart
Abestanis: Ah, thanks for the explanation. You are right, I wanted to use interfaces here, but I just forgot that `implements` exists for that so I landed on the mixin. 😄 Will change to an interface.

sweyer_plugin/lib/sweyer_plugin.dart
Abestanis: Hm, the problem is that these exceptions are app facing, the app wants to catch them and react to them.

So I guess the problem is that we are bringing method channel exceptions pretty much 1:1 into the app facing interface. Do you think defining separate exceptions in the `sweyer_plugin.dart` and then converting from method channel exceptions into those in the `sweyer_plugin_method_channel.dart` would be better?

sweyer_plugin/lib/sweyer_plugin.dart
Abestanis: Hm, this seems a bit unnecessary.
The abstract class with static methods is just an interface to the app and comes from the flutter federated plugin example. Its main job is to delegate the actual work to the `SweyerPluginPlatform` instance, which is where the DI can come in.

But this class is not meant to be instantiated and never has any state that would require an instance. In fact, maybe I should add a private constructor to prevent instantiation? Is there anything I'm overlooking here where the instance would be beneficial?

lib/routes/routes.dart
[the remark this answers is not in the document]
nt4f04uNd: It was previously working pretty fine with use `useInheritedMediaQuery: true`.

I see that the method I used to create the new `MediaQuery` is not entirely correct, but we could use `MediaQuery.fromWindow` instead.

So I would prefer that you return the `MediaQueryWrapper`, and do this `MediaQuery.fromWindow -> Builder -> MediaQuery(data: MediaQuery.of(context).copyWith(textScaleFactor: textScaleFactor), child: child)`.

I don't agree with mixing this into the router class.

sweyer_plugin/lib/sweyer_plugin.dart
nt4f04uNd: > The abstract class with static methods is just an interface to the app and comes from the flutter federated plugin example.

Yes, but abstract class with static methods is almost always a bad idea, because it doesn't have an interface, therefore it's not mockable, fakeable, testetable and is inherently a signletone (but worse).

In the past I rewrote the entirety of the app from this pattern just to make it testable https://github.com/nt4f04uNd/sweyer/pull/21

Let's not add additional tech-debt.

> But this class is not meant to be instantiated and never has any state that would require an instance.

Why not?

> Is there anything I'm overlooking here where the instance would be beneficial?

In the future everything should be in the DI, so classes don't have implicit dependencies through singletones or patterns like locator.

More of my thoughts on why this is bad is here https://github.com/nt4f04uNd/sweyer/issues/81

sweyer_plugin/lib/sweyer_plugin.dart
nt4f04uNd: > So I guess the problem is that we are bringing method channel exceptions pretty much 1:1 into the app facing interface. Do you think defining separate exceptions in the sweyer_plugin.dart and then converting from method channel exceptions into those in the sweyer_plugin_method_channel.dart would be better?

No, that's not a problem, the composition is fine.

I was only suggesting that you move it into the `sweyer_plugin_platform_interface.dart`, still use in the `sweyer_plugin_method_channel.dart` (there is no problem with that), and then export from the app-facing file.

The problem is that down layer (method channel) currently uses something from the upper layer (app-facing package), which shouldn't happen, it should be vice-versa.

lib/routes/routes.dart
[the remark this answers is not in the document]
nt4f04uNd: Okay, but let's move it into the MediaQueryWrapper

lib/routes/routes.dart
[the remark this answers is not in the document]
Abestanis: Ok, I didn't see `MediaQuery.fromWindow`, that is a good option.

lib/routes/routes.dart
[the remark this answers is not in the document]
Abestanis: Ok, we'll not have the `supportedLocales` in that case, I'll see if that breaks anything.

sweyer_plugin/lib/sweyer_plugin.dart
Abestanis: > Yes, but abstract class with static methods is almost always a bad idea, because it doesn't have an interface, therefore it's not mockable, fakeable, testetable and is inherently a signletone (but worse).

But the class is merely a wrapper around the `SweyerPluginPlatform`, which is the singleton in this case and already uses the pattern you are suggesting. I think that this class is a not singelton, it is merely a collection of static functions and therefore not more than a namespace (which is why it should never be instantiated).

Converting this namespace into a singleton which just delegates to another singleton is not worth it in my opinion. Because this class is just a wrapper, there is no use to mocking it when we can just mock the underlying `SweyerPluginPlatform`.

> In the future everything should be in the DI, so classes don't have implicit dependencies through singletones or patterns like locator.

I agree, we would be able to use `riverpod` in the static functions to resolve the `SweyerPluginPlatform`.

If you still think there is a need to mock the wrapping functions in the future, I'll convert the code to use the singleton pattern here. Let me know.

lib/routes/routes.dart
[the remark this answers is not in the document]
nt4f04uNd: > Ok, we'll not have the `supportedLocales` in that case, I'll see if that breaks anything.

Hm, I actually missed that.
The `supportedLocales` part itself shoudn't break anything, but the absense of the app context can.

For example, I can see that `reportError` calls `ShowFunctions.instance.showError`, which would throw without the app being rendered.

Let's keep it inside the Router, it seems like a safer variant.

sweyer_plugin/lib/sweyer_plugin.dart
nt4f04uNd: > But the class is merely a wrapper around the `SweyerPluginPlatform`, which is the singleton in this case and already uses the pattern you are suggesting. I think that this class is a not singelton, it is merely a collection of static functions and therefore not more than a namespace (which is why it should never be instantiated).

The user shouldn't know or care about it.

Ideally we shouldn't have been faking the `FakeSweyerPluginPlatform`, but `SweyerPlugin`, but it's fine to leave as-is right now.

> If you still think there is a need to mock the wrapping functions in the future, I'll convert the code to use the singleton pattern here. Let me know.

Yes, I still think this would be valuable.

sweyer_plugin/lib/sweyer_plugin.dart
Abestanis: You are right, it makes more sense that way. I still felt decoupling the `PlatformException` a bit more from the `SweyerExeption` would make the interface a bit nicer. Let me know if that is ok for you.

lib/widgets/content_art.dart
```diff
@@ -430,14 +430,12 @@ class _SongScopedStorageArtSourceLoader extends _ArtSourceLoader {
...
-      } on SweyerPluginException catch (ex, stack) {
-        if (ex != SweyerPluginException.io) {
-          FirebaseCrashlytics.instance.recordError(
-            ex,
-            stack,
-            reason: 'in _SongScopedStorageArtSourceLoader.load',
-          );
-        }
+      } on SweyerPluginIoException catch (ex, stack) {
+        FirebaseCrashlytics.instance.recordError(
```
nt4f04uNd, line 434: This is now doing exactly the opposite of what is supposed to do.

We specifically to record don't want IO exceptions here, because they are caused by non-existing art (which is not our fault the system provided to us) and by canceling out the loading.

lib/logic/player/content.dart
```diff
@@ -740,11 +740,8 @@ class ContentControl extends Control {
       await SweyerPlugin.renamePlaylist(playlist, name);
       await refetchSongsAndPlaylists();
       return name;
-    } on SweyerPluginException catch (ex) {
-      if (ex == SweyerPluginException.playlistNotExists) {
-        return null;
-      }
-      rethrow;
+    } on PlaylistNotExistException catch (_) {
```
nt4f04uNd, line 743: `catch (_)` can be omitted

sweyer_plugin/lib/sweyer_plugin_method_channel.dart
```diff
@@ -7,41 +7,25 @@ import 'package:flutter/services.dart';
...
-
+enum SweyerMethodChannelExceptionCode {
   /// Generic error.
-  static const unexpected = SweyerMethodChannelException._('UNEXPECTED_ERROR');
+  unexpected('UNEXPECTED_ERROR'),
 
   /// On Android 30 requests like `MediaStore.createDeletionRequest` require
   /// calling `startIntentSenderForResult`, which might throw this exception.
-  static const intentSender = SweyerMethodChannelException._('INTENT_SENDER_ERROR');
+  intentSender('INTENT_SENDER_ERROR'),
```
nt4f04uNd, line 16: where did the `intentSender` come?

sweyer_plugin/lib/sweyer_plugin_platform_interface.dart
```diff
@@ -5,6 +5,44 @@ import 'package:plugin_platform_interface/plugin_platform_interface.dart';
 
 import 'sweyer_plugin_method_channel.dart';
 
+
+/// Base class for all exceptions from the Sweyer plugin.
+class SweyerPluginException implements Exception {
+  /// The exception reason.
+  final String message;
+  /// An optional cause of this exception.
+  final Exception? cause;
```
nt4f04uNd, line 14: Why is it optional?

sweyer_plugin/lib/sweyer_plugin_method_channel.dart
```diff
@@ -152,8 +132,11 @@ class MethodChannelSweyerPlugin extends SweyerPluginPlatform {
           'name': name,
         },
       );
-    } on PlatformException catch (ex) {
-      throw SweyerMethodChannelException._throw(ex, const [SweyerMethodChannelException.playlistNotExists]);
+    } on PlatformException catch (error) {
+      if (error.code == SweyerMethodChannelExceptionCode.playlistNotExists.value) {
+        throw PlaylistNotExistException(playlistId, cause: error);
```
nt4f04uNd, line 137: What is the differentiation between "common" and not common? Why `_convertCommonExceptions` doesn't just handle all of them?

Why is this method not a part of the platform interface?

lib/widgets/content_art.dart
Abestanis: Ah, sorry, you are right.

sweyer_plugin/lib/sweyer_plugin_method_channel.dart
Abestanis: We were never specifically handling it, so it is just a generic `SweyerException` now. But it is captured in the `cause`.

sweyer_plugin/lib/sweyer_plugin_platform_interface.dart
Abestanis: We might want to throw other exceptions in the future without having a causing exception.

sweyer_plugin/lib/sweyer_plugin_method_channel.dart
nt4f04uNd: > We were never specifically handling it, so it is just a generic `SweyerException` now. But it is captured in the `cause`.

I think if we are doing this, all exception codes should have a separate exception for them

sweyer_plugin/lib/sweyer_plugin_method_channel.dart
Abestanis: > What is the differentiation between "common" and not common?

Well, the "common" exceptions don't have a parameter like the `playlistId`. Please suggest a better name if you have one, I'm also not too thrilled with the current one.

>  Why _convertCommonExceptions doesn't just handle all of them?

Do you think it would be better to pass the `playlistId` as an optional parameter? This might get ugly in the future, if we add other errors with parameters.

> Why is this method not a part of the platform interface?

Because this converts `PlatformException`s, which are specific to method channels. A mock implementation of the `SweyerPluginPlatform` interface or any other theoretical implementation that does not use method channels does not need this method.

nt4f04uNd: Ah, I see now, thanks

sweyer_plugin/lib/sweyer_plugin_method_channel.dart
Abestanis: The indent sender is very Android specific and not very descriptive, I decided to name the error `PlatformHandlerException` instead. Maybe I would need something like that on Windows as well. I hope that's ok for you.

nt4f04uNd:
Thanks! 🎉

PR #90, opened 2023-04-30
Fix settings button visible on insertion settings page

Abestanis:
This was originally fixed by #50, but seems to have regressed since then, probably due to a flutter update.

Looks like the dependency on `settingsOpened` was no longer registered in the `initState` function. We work around that by using a `StatefulBuilder`.

It is a bit weird that the goldens on CI did not pick this up in #85, they were showing the button consistently locally.
To make sure this doesn't regress in the future, I also added some tests for the selection route.

### DIFF
--- a/test/fakes/fake_sweyer_plugin_platform.dart
+++ b/test/fakes/fake_sweyer_plugin_platform.dart
@@ -61,7 +61,7 @@ class FakeSweyerPluginPlatform extends SweyerPluginPlatform {
     required int playlistId,
   }) async {
     final playlist = playlists!.firstWhere((playlist) => playlist.id == playlistId);
-    playlist.songIds.insertAll(index - 1, songIds);
+    playlist.songIds.insertAll(index, songIds);
   }
 
   @override
--- a/test/fakes/fake_sweyer_plugin_platform.dart
+++ b/test/fakes/fake_sweyer_plugin_platform.dart
@@ -59,7 +59,10 @@ class FakeSweyerPluginPlatform extends SweyerPluginPlatform {
     required int index,
     required List<int> songIds,
     required int playlistId,
-  }) async {}
+  }) async {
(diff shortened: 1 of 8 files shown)


test/fakes/fake_sweyer_plugin_platform.dart
```diff
@@ -58,71 +57,80 @@ class FakeContentChannel implements ContentChannel {
...
   Future<void> insertSongsInPlaylist({
     required int index,
-    required List<Song> songs,
-    required Playlist playlist,
-  }) async {}
+    required List<int> songIds,
+    required int playlistId,
+  }) async {
+    final playlist = playlists!.firstWhere((playlist) => playlist.id == playlistId);
+    playlist.songIds.insertAll(index - 1, songIds);
```
Abestanis, line 64: I was a bit surprised that the index is not zero based, but this seems by design:
https://github.com/nt4f04uNd/sweyer/blob/70711b3b2015fd4dce0fb5eea446ce0f01131b84/lib/routes/home_route/persistent_queue_route.dart#L244-L256

Are you sure this is correct? I could not find any more info about the [`PLAY_ORDER`](https://developer.android.com/reference/android/provider/MediaStore.Audio.Playlists.Members#PLAY_ORDER) entry.

nt4f04uNd: Yes, I'm sure. Both 0 and 1 give the same result.

nt4f04uNd: You can leave a comment about this here or the function signature

Abestanis: Since a one based index is very unintuitive and we are going to move away from using the Mediastore for playlists anyways, I changed `insertSongsInPlaylist` to take a 0 based index and add one to it on the Android native side.

nt4f04uNd (approved):

PR #91, opened 2023-04-30
Decide whether to show scroll labels based on drag position

### LINKED ISSUE #88: Use drag position instead of scrollbar thumb to determine whether to show labels
### Suggestion

When the user drags the scroll bar of the content lists (Tracks, Albums, Playlists and Artists) we show a label if the sorting is set to name or title.

Scroll label

This is a cool feature, but it is very hard to get to, because the scroll bar is only a few pixels wide. Additionally, to provide this feature we are keeping a copy of the Scrollbar code from Flutter, which requires us to manually keep it up to date with the Flutter implementation to get improvements and to keep compatible with changes in the framework.

I want to get your feedback on an alternative solution to this problem that does not require us to keep a slightly modified version of the code and makes the feature easier to access:

Instead of only showing the label when the scrollbar is dragged, show it when a scrolling from a drag, while the touchpoint of the drag is close to the window border on the scrollbar side. For the picture above, this would mean when dragging the list close to the right side of the screen, the labels would show up.

We can implement this using a `NotificationListener<ScrollNotification>` and the `dragDetails?.localPosition` of `ScrollStartNotification` & `ScrollUpdateNotification`.

I would like to implement this, if you are ok with it, @nt4f04uNd.

Abestanis:
Instead of detecting the drag of the scroll bar, allow any drag that is close to the scrollbar side of the window to show the scroll labels. Fixes #88.

New scroll label behaviour

Should we show the labels during overscroll? I currently have it enabled, but in the video below you can see that we get a label even if we don't have any actual scrolling because of this:

<details>
<summary>Scrolling without scrolling</summary>

Scroll indicators show without actually scrolling
</details>


lib/widgets/scrollbar.dart
```diff
@@ -1,29 +1,12 @@
 /*---------------------------------------------------------------------------------------------
 *  Copyright (c) nt4f04und. All rights reserved.
 *  Licensed under the BSD-style license. See LICENSE in the project root for license information.
-*
-*  Copyright (c) The Flutter Authors.
-*  See ThirdPartyNotices.txt in the project root for license information.
 *--------------------------------------------------------------------------------------------*/
```
Abestanis, line 4: Do we still want the copyright header?

nt4f04uNd: No, it's no longer needed - I only put it in files that contain code from external sources

test/golden/routes_golden_test.dart
```diff
@@ -128,6 +128,27 @@ void main() {
...
+      });
+      await tester.runAppTest(() async {
+        ContentControl.instance.sort(
+            contentType: ContentType.song, sort: const SongSort(feature: SongSortFeature.title));
+        await tester.pumpAndSettle();
+
+        final listRect = tester.getRect(find.byType(ContentListView).hitTestable());
+        final gesture = await tester.startGesture(Offset(listRect.right - 20, listRect.bottom - 20));
+        for (int i = 0; i < 10; i++) {
+          await tester.pump(const Duration(milliseconds: 50));
```
nt4f04uNd, line 145: There's a `tester.timedDragFrom` for it, but this is acceptable too

lib/widgets/scrollbar.dart
```diff
@@ -122,474 +109,69 @@ class _AppScrollbarState extends State<AppScrollbar> {
...
+            } else if (notification is ScrollUpdateNotification) {
+              dragPosition = notification.dragDetails?.localPosition;
+            } else if (notification is OverscrollNotification) {
+              dragPosition = notification.dragDetails?.localPosition;
+            } else {
+              dragPosition = null;
+            }
+            final windowSize = context.size;
+            setState(() {
+              showScrollLabels = windowSize != null &&
```
nt4f04uNd, line 133: This setState is too often.
We don't actually need to call it unless the value has changed.

lib/widgets/scrollbar.dart
```diff
@@ -122,474 +109,69 @@ class _AppScrollbarState extends State<AppScrollbar> {
...
-              }),
-              onThumbPressEnd: () => setState(() {
-                _dragIsActive = false;
-              }),
-              child: Theme(
-                data: theme.copyWith(highlightColor: highlightColor),
-                child: widget.child,
+        child: NotificationListener<ScrollNotification>(
+          onNotification: (notification) {
+            final Offset? dragPosition;
```
nt4f04uNd, line 121: We don't need to handle any of that when `labelBuilder == null`

nt4f04uNd (commented):
Ahh, I actually thought you want to add a fully featured alphabetical scroll. But it's still a good addition 👍

test/golden/routes_golden_test.dart
Abestanis: Unfortunately, we can't use it because it releases the touch (does a pointer up event). We need the pointer/finger to be down when taking the golden image.

lib/widgets/scrollbar.dart
```diff
@@ -116,61 +142,36 @@ class _AppScrollbarState extends State<AppScrollbar> {
...
-                notificationPredicate: widget.notificationPredicate,
-                interactive: widget.interactive,
-                child: Theme(
-                  data: theme.copyWith(highlightColor: highlightColor),
-                  child: widget.child,
-                ),
+        child: widget.labelBuilder == null
+            ? scrollbarChild
+            : NotificationListener<ScrollNotification>(
+                onNotification: (notification) {
```
nt4f04uNd, line 148: I'd prefer if the tree didn't change when this property changes, since this means that the children State's will be lost.

We can just return false from the notification listener.

Abestanis: Ah, yeah, that's better 👍

nt4f04uNd (approved):

PR #95, opened 2023-05-01
Improve startup robustness

Abestanis:
This fixes a few problem that would cause the app to get stuck on the splash screen, when there was an error during startup.
I think it is a better user experience when we continue and show the app UI, even though some content might be missing because of the exception.

### DIFF
--- a/lib/logic/player/content.dart
+++ b/lib/logic/player/content.dart
@@ -437,6 +437,7 @@ class ContentControl extends Control {
         stack,
         reason: 'in re-fetch ${contentType.name}',
       );
+      ShowFunctions.instance.showToast(msg: staticl10n.oopsErrorOccurred);
     }
   }
 
--- a/lib/logic/player/content.dart
+++ b/lib/logic/player/content.dart
@@ -381,54 +381,62 @@ class ContentControl extends Control {
     if (disposed.value) {
       return;
     }
-    switch (contentType) {
-      case ContentType.song:
-        state.allSongs.setSongs((await SweyerPlugin.instance.retrieveSongs(Song.fromMap)).toList());
-        if (_empty) {
-          dispose();
-          return;
-        }
-        sort(emitChangeEvent: false, contentType: contentType);
-        if (updateQueues) {
-          QueueControl.instance.removeObsolete(emitChangeEvent: false);
-        }
-        break;
-      case ContentType.album:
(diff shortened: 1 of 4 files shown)


sweyer_plugin/lib/sweyer_plugin_method_channel.dart
```diff
@@ -138,8 +168,13 @@ class MethodChannelSweyerPlugin extends SweyerPluginPlatform {
   }
 
   @override
-  Future<void> removePlaylists(List<int> playlistIds) async =>
-      methodChannel.invokeMethod<void>('removePlaylists', {'ids': playlistIds});
+  Future<void> removePlaylists(List<int> playlistIds) async {
+    try {
+      return methodChannel.invokeMethod<void>('removePlaylists', {'ids': playlistIds});
```
nt4f04uNd, line 173: this will be not catched without `await`

lib/logic/player/content.dart
```diff
@@ -381,54 +381,62 @@ class ContentControl extends Control {
...
+            QueueControl.instance.resetQueueAsFallback();
+          }
+          sort(emitChangeEvent: false, contentType: contentType);
+          break;
+      }
+      if (emitChangeEvent) {
+        emitContentChange();
+      }
+    } on SweyerPluginException catch (ex, stack) {
+      FirebaseCrashlytics.instance.recordError(
```
nt4f04uNd, line 435: I don't think this method should catch exceptions.
Exception here means something has gone incredibly wrong.

What will user see if on start up this has thrown an error?

Also, this method is called from more places than just `init`. It's not a responsibility of this method to decide what to do with the error.

nt4f04uNd: Actually, I think it **can** catch the error, but it should then show a toast to a user that something went wrong.

sweyer_plugin/lib/sweyer_plugin_method_channel.dart
Abestanis: Good catch, thanks

lib/logic/player/content.dart
Abestanis: Yeah, I think we must catch the error here or check every call site and make sure to catch the error there. Because during the conversion from Java to Android I did have some instances where the native side threw an error and the App would just be stuck in the launch screen, which is not a good user experience.

But I do agree that we need to show something if possible, so let's add a toast.

nt4f04uNd (approved):

PR #120, opened 2024-05-24
Fix running app in a different zone than the framework was initialized

Abestanis:
When the app starts up with the latest Flutter version it throws this <details><summary>error</summary>

```
======== Exception caught by Flutter framework =====================================================
The following assertion was thrown during runApp:
Zone mismatch.

The Flutter bindings were initialized in a different zone than is now being used. This will likely cause confusion and bugs as any zone-specific configuration will inconsistently use the configuration of the original binding initialization zone or this zone based on hard-to-predict factors such as which zone was active when a particular callback was set.
It is important to use the same zone when calling `ensureInitialized` on the binding as when calling `runApp` later.
To make this warning fatal, set BindingBase.debugZoneErrorsAreFatal to true before the bindings are initialized (i.e. as the first statement in `void main() { }`).
When the exception was thrown, this was the stack:
#0      BindingBase.debugCheckZone.<anonymous closure> (package:flutter/src/foundation/binding.dart:495:29)
#1      BindingBase.debugCheckZone (package:flutter/src/foundation/binding.dart:500:6)
#2      _runWidget (package:flutter/src/widgets/binding.dart:1439:18)
#3      runApp (package:flutter/src/widgets/binding.dart:1373:3)
#4      main.<anonymous closure> (package:sweyer/main.dart:104:5)
<asynchronous suspension>
====================================================================================================
```
</details>

~This is fixed by ensuring that the `WidgetBindings` are initialized in the zone created by `runZonedGuarded`. That required to improve the `reportError` function to handle getting called before anything is initialized.~

I realized that there is a better fix for this, which is to remove the `runZonedGuarded` and use `PlatformDispatcher.instance.onError` instead. This is the recommended way to catch async errors per [the crashlytics documentation](https://firebase.google.com/docs/crashlytics/customize-crash-reports?platform=flutter#asynchronous-errors) for the latest version.

To make sure everything continues to work I've updated the firebase packages to the latest version. They added an autogenerated `firebase_options.dart`, which I generated using a test firebase project and copying the values from the `google-services.json` file. It would be best if you could run `flutterfire configure` as described in [the documentation](https://firebase.google.com/docs/crashlytics/get-started?platform=flutter#add-sdk) to make sure the values are correct.


lib/main.dart
```diff
@@ -20,15 +20,12 @@ $stack''';
 }
 
 Future<void> reportError(dynamic ex, StackTrace stack) async {
-  if (Prefs.devMode.get()) {
-    ShowFunctions.instance.showError(
-      errorDetails: buildErrorReport(ex, stack),
-    );
+  if (Prefs.isInitialized && Prefs.devMode.get()) {
```
Abestanis, line 23:
```suggestion
  if (kDebugMode || (Prefs.isInitialized && Prefs.devMode.get())) {
```
Should we maybe just always show the error toast if we were build in debug mode?

android/app/build.gradle
```diff
@@ -1,3 +1,14 @@
+plugins {
```
Abestanis, line 1: The new flutter version changed the android gradle project structure a bit. These changes are only the ones related to firebase cherry picked from all changes for the flutter upgrade. I unfortunately can't test if the app can compile with just these changes. ☹️

lib/main.dart
```diff
@@ -19,16 +20,15 @@ $ex
...
-      errorDetails: buildErrorReport(ex, stack),
+      errorDetails: buildErrorReport(error, stack),
     );
   }
-  await FirebaseCrashlytics.instance.recordError(
-    ex,
-    stack,
-  );
+  if (Firebase.apps.isNotEmpty) {
+    await FirebaseCrashlytics.instance.recordError(error, stack, fatal: true);
```
Abestanis, line 30: We need to mark errors as fatal now, otherwise they don't get reported ([docs](https://firebase.google.com/docs/crashlytics/customize-crash-reports?platform=flutter#report-caught-exceptions)):
Firebase documentation for non-fatal errors

lib/main.dart
nt4f04uNd: It was a bit annoying in the past, because it could generate subsequent errors from the original one

lib/main.dart
nt4f04uNd: Ok, but this is a weird move from them

lib/main.dart
```diff
@@ -66,38 +69,40 @@ class _WidgetsBindingObserver extends WidgetsBindingObserver {
...
+    return true;
+  };
+  WidgetsBinding.instance.addObserver(_WidgetsBindingObserver());
+
+  await DeviceInfoControl.instance.init();
+  ThemeControl.instance.init();
+  ThemeControl.instance.initSystemUi();
+  await Permissions.instance.init();
+  await ContentControl.instance.init();
+  runApp(const ProviderScope(child: App()));
```
nt4f04uNd, line 105: Should we also do the ` FirebaseCrashlytics.instance.recordFlutterFatalError` parat from the [docs](https://firebase.google.com/docs/crashlytics/customize-crash-reports?platform=flutter#asynchronous-errors) ?

android/app/build.gradle
nt4f04uNd: We should not merge this until we can be sure it does

nt4f04uNd:
> To make sure everything continues to work I've updated the firebase packages to the latest version. They added an autogenerated `firebase_options.dart`, which I generated using a test firebase project and copying the values from the `google-services.json` file. It would be best if you could run `flutterfire configure` as described in [the documentation](https://firebase.google.com/docs/crashlytics/get-started?platform=flutter#add-sdk) to make sure the values are correct.

I've done and it only generated minor formatting changes, it seems

lib/main.dart
Abestanis: Agreed.

Abestanis:
> I've done and it only generated minor formatting changes, it seems

Thanks 👍

android/app/build.gradle
Abestanis: I hope #121 will help with this.

lib/main.dart
Abestanis: Yeah, I guess it's better to leave it as an explicit setting. 👍

android/app/build.gradle
Abestanis: The flutter version we use is too old to handle the new way of defining the plugins, so I have to revert this:

CI failure

nt4f04uNd (approved):

nt4f04uNd:
New build workflow is amazing to manually test PRs

PR #118, opened 2024-03-24
Fix Fastlane CD

nt4f04uNd:
Follow up after https://github.com/nt4f04uNd/sweyer/pull/117

PR #121, opened 2024-05-26
Add CI job to build APK

Abestanis:
It would be good if CI would notify us if a change would break building the APK.

Unfortunately this increases our CI runtime from ~2min 30sec to ~4min 30sec. I already added a cache for Gradle, which helped a ton, but caching the Android SDK is not helping.

### DIFF
--- a/.github/workflows/build_and_test/action.yml
+++ b/.github/workflows/build_and_test/action.yml
@@ -47,19 +47,6 @@ runs:
         path: test/golden/failures
         if-no-files-found: ignore
 
-    - name: 💾 Cache Android SDK
-      uses: actions/cache@v4
-      with:
-        key: android-sdk-${{inputs.flutter_version}}-${{inputs.flutter_channel}}
-        path: |
-          /home/runner/.android
-          /usr/local/lib/android/sdk/emulator
-          /usr/local/lib/android/sdk/platform-tools
-          /usr/local/lib/android/sdk/tools
-          /usr/local/lib/android/sdk/build-tools/30.0.3
-          /usr/local/lib/android/sdk/platforms/android-30
-          /usr/local/lib/android/sdk/platforms/android-28
-
     - name: 💾 Cache Gradle
       uses: actions/cache@v4
       with:
@@ -68,7 +55,4 @@ runs:
 
     - name: 🔨 Build Apk
       shell: bash
-      run: |
-        ls /usr/local/lib/android/sdk/licenses
-        flutter build apk --debug
-        ls /usr/local/lib/android/sdk/licenses
+      run: flutter build apk --debug
--- a/.github/workflows/build_and_test/action.yml
+++ b/.github/workflows/build_and_test/action.yml
@@ -47,6 +47,15 @@ runs:
         path: test/golden/failures
         if-no-files-found: ignore
 
+    - name: 💾 Cache android SDK
+      uses: actions/cache@v2
+      id: android-sdk
+      with:
+        key: android-sdk
+        path: |
+          /usr/local/lib/android/adk
+          ~/.android
+
     - name: 🔨 Build Apk
       shell: bash
       run: flutter build apk --debug
--- a/.github/workflows/build_and_test/action.yml
+++ b/.github/workflows/build_and_test/action.yml
@@ -35,10 +35,6 @@ runs:
         very_good --analytics false
         very_good packages get --recursive
 
-    - name: 🔨 Build Apk
-      shell: bash
-      run: flutter build apk --debug
-
     - name: 🧪 Run Tests
       shell: bash
       run: flutter test --no-pub --test-randomize-ordering-seed random ${{inputs.testing_arguments}}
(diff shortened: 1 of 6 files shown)


Abestanis:
What do you think of adding a label and only running the new steps if the label is applied to the merge request?

nt4f04uNd:
> What do you think of adding a label and only running the new steps if the label is applied to the merge request?

I think it's okay to make it run on all PRs

.github/workflows/build_and_test/action.yml
```diff
@@ -46,3 +46,13 @@ runs:
...
+
+    - name: 💾 Cache Gradle
+      uses: actions/cache@v4
+      with:
+        key: gradle
+        path: /home/runner/.gradle
+
+    - name: 🔨 Build Apk
+      shell: bash
+      run: flutter build apk --debug
```
nt4f04uNd, line 58: Should it build profile instead?

nt4f04uNd, line 58: Let's go one step further and upload it as an artifact

nt4f04uNd:
> Unfortunately this increases our CI runtime from ~2min 30sec to ~4min 30sec. I already added a cache for Gradle, which helped a ton, but caching the Android SDK is not helping.

Can we just make it a separate job from tests?

.github/workflows/build_and_test/action.yml
nt4f04uNd: Though, I suppose it would require a release signature, which is currently WIP in https://github.com/nt4f04uNd/sweyer/pull/118/files

nt4f04uNd: I think we could leave it debug for now, and create an issue to make it profile

Abestanis: I chose to use a debug build to avoid requiring any signing keys, and to avoid spending time on minification.

Since the point of this is mainly to proof that the apk compiles, I'm not sure what benefit switching to profile would bring. Could you please elaborate on why it would be better?

Abestanis: Ah yeah, good idea 👍

Abestanis:
> Can we just make it a separate job from tests?

Of course, that is a good solution! We can just make it a workflow that is triggered manually like the update_goldens workflow.

Abestanis:
Looks good, see [test run](https://github.com/Abestanis/sweyer/actions/runs/9489936411)

New actions page

Note: I went with `executables` instead of `apk` as the name and description of the workflow, because I still want to make Sweyer run on desktop at some point and then we should also build the other binaries in that workflow.

.github/workflows/build_and_test/action.yml
nt4f04uNd: > Since the point of this is mainly to proof that the apk compiles, I'm not sure what benefit switching to profile would bring. Could you please elaborate on why it would be better?

It's useful to manually test the app from the PR. It gives you the closes build to production, while still leaving some of the debug tooling

.github/workflows/build_executables.yml
```diff
@@ -0,0 +1,40 @@
+name: Build Sweyer executables
+on: 
+  workflow_dispatch
```
nt4f04uNd, line 3: Let's make it run on each PR and make it required

If some PR breaks app build, it should not be merged

nt4f04uNd:
Let's rename[ this workflow](https://github.com/nt4f04uNd/sweyer/blob/fd715b7b7245ef7dfed9e289a4ff589da7cb8aa9/.github/workflows/build_and_test/action.yml#L5) and remove the part about building Sweyer, it's a bit misleading

Abestanis:
> Let's rename[ this workflow](https://github.com/nt4f04uNd/sweyer/blob/fd715b7b7245ef7dfed9e289a4ff589da7cb8aa9/.github/workflows/build_and_test/action.yml#L5) and remove the part about building Sweyer, it's a bit misleading

Ok, should we rename the workflow to `test` and change the description to just `Test Sweyer`?

.github/workflows/build_executables.yml
Abestanis: Ok, done👍

.github/workflows/build_and_test/action.yml
Abestanis: Ah, I didn't consider that we would want to use it for manual testing, I just wanted a way to verify that the apk can be build. This makes sense, I changed it to profile.

nt4f04uNd:
> Ok, should we rename the workflow to `test` and change the description to just `Test Sweyer`?

SGTM

.github/workflows/build_executables.yml
```diff
@@ -31,10 +31,10 @@ jobs:
 
       - name: 🔨 Build Apk
         shell: bash
-        run: flutter build apk --debug
+        run: flutter build apk --profile
```
nt4f04uNd, line 34: I'm surprised. Does not need release signatures?

nt4f04uNd (approved):

.github/workflows/build_executables.yml
Abestanis: We only configured signing for `release`, `assembleProfile` doesn't seem to use the signing config:

https://github.com/nt4f04uNd/sweyer/blob/b96a5ee3eead62e546059a13b4071c59f7a11454/android/app/build.gradle#L60-L61

Issue #122, opened 2024-06-15
Hangs on startup due to missing album metadata


Abestanis:
### Steps to reproduce

0. Have some music with missing metadata. Probably a missing year, but I haven't found out which song is the culprit.

Either
1. Update to the latest version.
2. Start the app.
3. Get stuck on the splash screen forever.

Or
1. Freshly install the latest version.
2. Open the app.
3. Grant permissions.
4. Get stuck forever searching for music.

### What did you expect?

For the app to start.

### What happened instead?

We just hang forever because we throw an exception here
https://github.com/nt4f04uNd/sweyer/blob/7496517cb5572b1e00ce3ac572bbde2a62e30b11/lib/logic/models/album.dart#L103
and here
https://github.com/nt4f04uNd/sweyer/blob/7496517cb5572b1e00ce3ac572bbde2a62e30b11/lib/logic/models/album.dart#L104

The value for `fistYear` and `lastYear` seem to be `null` for at least one album, so the cast to `int` fails.

We should:
* Change these values to be nullable (`lastYear` already is, we just need to change the cast).
* Be more robust and don't fail the entire content search because one item fails. We should catch exceptions on an item level and ignore them.
* Catch errors in the content search and show an error UI instead of hanging forever.

I would love to submit a MR so we can patch this as soon as possible (I think this deserves a hot fix release), but I'm on vacation for a week so I won't be able to work on it until I'm back.

### Sweyer version

1.0.10

### Android version

14

Abestanis:
Also, I'm wondering if you saw these exceptions in Fire ase.

nt4f04uNd:
@Abestanis I added you as a viewer to the Firebase project

nt4f04uNd:
I do see a few exceptions there

nt4f04uNd:
I uploaded the debug symbols, and new crashes should automatically be deobfuscated for this version

Made an issue to automate it
https://github.com/nt4f04uNd/sweyer/issues/123

nt4f04uNd:
Deobfuscated stacktrace

```
          Non-fatal Exception: io.flutter.plugins.firebase.crashlytics.FlutterError: type 'Null' is not a subtype of type 'int' in type cast
#00 pc 0x3fe127 com.nt4f04und.sweyer (new Album.fromMap [album.dart:103]) (BuildId: bca64abfc060cd5531bb8f1c719e3625)
[...]

PR #124, opened 2024-06-22
Fix firstYear crash

nt4f04uNd:
Fixes https://github.com/nt4f04uNd/sweyer/issues/122

### DIFF
--- a/lib/logic/models/album.dart
+++ b/lib/logic/models/album.dart
@@ -11,7 +11,7 @@ class Album extends PersistentQueue implements PlatformAlbum {
   final String? albumArt;
   final String artist;
   final int? artistId;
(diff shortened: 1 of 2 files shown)


Abestanis (approved):
Looks good 👍

Abestanis:
Sorry, I was too quick there, its hard to review on the phone. Shouldn't it also be nullable in`_AlbumCopyWith`?

nt4f04uNd:
> Sorry, I was too quick there, its hard to review on the phone. Shouldn't it also be nullable in`_AlbumCopyWith`?

It's okay, this code is not called anywhere anyway. You are right, it should be. We should probably make a codegen one day...

PR #130, opened 2024-06-30
Update Flutter version

Abestanis:
Update Flutter to `3.22.2` and also update dependencies.

Depends on https://github.com/nt4f04uNd/nt4f04unds_widgets/pull/12, https://github.com/nt4f04uNd/FlutterToast/pull/1 and https://github.com/nt4f04uNd/audio_service/pull/1.


.github/workflows/build_executables.yml
```diff
@@ -12,15 +12,14 @@ jobs:
...
-          flutter-version: 3.7.12
+          flutter-version: 3.22.2
           channel: stable
           cache: true
 
       - name: 📦 Install Dependencies
         shell: bash
         run: |
           flutter pub global activate very_good_cli
-          very_good --analytics false
```
Abestanis, line 23: See https://github.com/VeryGoodOpenSource/very_good_cli/pull/756.

lib/constants/theme/themes.dart
```diff
@@ -13,6 +13,7 @@ abstract class Theme {
 
   static ThemeContainer<ThemeData> app = ThemeContainer(
     light: ThemeData(
+      useMaterial3: false,
```
Abestanis, line 16: Flutter 3.16.0 switched the default value of `useMaterial3` to `true` (see [PR](https://github.com/flutter/flutter/pull/129724)). Eventually, we want to switch to Material3, but it changes the theming and the way text themes are inherited, which causes layout and theme changes in the app (see [goldens diffs from before setting the property back to `false`](https://github.com/Abestanis/sweyer/commit/3119509e9135d72e9b82491bcf241f64a096f5a3)). Because I want to keep the diff small, I decided to leave switching to Material3 to a dedicated pull request.

test/test.dart
```diff
@@ -307,41 +307,37 @@ void testAppGoldens(
...
+        lightTheme: lightTheme,
+        playerInterfaceColorStyle: nonDefaultPlayerInterfaceColorStyle ? playerInterfaceColorStyle : null,
+      );
+
+      testGoldens(
+        testDescription.buildDescription(description),
+        (tester) async {
+          final previousDeterministicCursor = EditableText.debugDeterministicCursor;
+          addTearDown(() {
+            EditableText.debugDeterministicCursor = previousDeterministicCursor;
```
Abestanis, line 325: Moving the `EditableText.debugDeterministicCursor = previousDeterministicCursor;` into a `addTearDown` callback fixes the cursor not showing up in golden screenshots. My guess is that these screenshots are generated asynchronously and now take longer than before, even after `testGoldens` returns, so resetting `debugDeterministicCursor` synchronously was resetting it too early.

Abestanis: This approach was stolen from the [Flutter golden tests](https://github.com/flutter/flutter/blob/a3c7094d7ddae7ff49c56dd4f32dc458d10e4342/packages/flutter/test/widgets/editable_text_cursor_test.dart#L1145).

Abestanis:
@nt4f04uNd The new version `0.4.2` of `android_content_provider` seems to be a breaking the build on pre Flutter `3.24` (see [the failed CI](https://github.com/nt4f04uNd/sweyer/actions/runs/10435325222/job/28899161030?pr=130)).

It might be advantageous to retract the `0.4.2` version and instead release a `0.5.0` version, maybe even add a dependency on the Flutter version if possible.

Abestanis:
This is now only blocked by https://github.com/nt4f04uNd/audio_service/pull/1.

Abestanis:
The Codecov status checks don't seem to be showing up now even though the upload succeeds and no errors are shown in the CI. @nt4f04uNd when you have some time could you check if we have the app enabled for this repository and maybe disable and re-enable it as suggested [here](https://community.codecov.com/t/github-checks-do-not-show-up/3647/3).

nt4f04uNd:
> The Codecov status checks don't seem to be showing up now even though the upload succeeds and no errors are shown in the CI. @nt4f04uNd when you have some time could you check if we have the app enabled for this repository and maybe disable and re-enable it as suggested [here](https://community.codecov.com/t/github-checks-do-not-show-up/3647/3).

I didn't have it installed, should be working now

Abestanis:
> I didn't have it installed, should be working now

Yes, looks like it's working now, thanks!

nt4f04uNd:
> @nt4f04uNd The new version `0.4.2` of `android_content_provider` seems to be a breaking the build on pre Flutter `3.24` (see [the failed CI](https://github.com/nt4f04uNd/sweyer/actions/runs/10435325222/job/28899161030?pr=130)).
> 
> It might make sense to retract the `0.4.2` version and instead release a `0.5.0` version, maybe even add a dependency constraint on the Flutter version if possible.

Will check that https://github.com/nt4f04uNd/android_content_provider/issues/21

.gitignore
```diff
@@ -23,62 +24,24 @@
...
-# Android related
-**/android/**/gradle-wrapper.jar
-**/android/.gradle
-**/android/captures/
-**/android/gradlew
-**/android/gradlew.bat
-**/android/local.properties
-**/android/**/GeneratedPluginRegistrant.java
-
-# iOS/XCode related
```
nt4f04uNd, line 43: Why touch this? I would definitely like to preserve iOS ignores

But other changes to this file are questionable as well

UPD: apparently it's because of the generated `android/.gitignore`. I would rather have a unified .gitignore file

lib/constants/theme/themes.dart
nt4f04uNd: > Because I want to keep the diff small, I decided to leave switching to Material3 to a dedicated pull request.

I am not sure if we even want to do this, the whole app design is currenly more about material 2 (and it seems OK honestly)

android/app/src/main/AndroidManifest.xml
```diff
@@ -1,5 +1,5 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
-    package="com.nt4f04und.sweyer">
```
nt4f04uNd, line 2: It is really not needed?
upd: checked, docs conifrm it's needed to remove it https://developer.android.com/guide/topics/manifest/manifest-element#package

.gitignore
Abestanis: My workflow when updating the project is like this:
* I create a new empty Flutter project.
* I do a directory comparison between our project and the newly generated project.
* I manually go through the diff and check what should be changed.

The new flutter project template splits the unified `.gitignore` and moves the Android specific ignores into `android/.gitignore` and the IOS related ones into `ios/.gitignore`. I'm trying to bring our project as close as possible to the Flutter project template, because it reduces the noise of our own changes in the diff and therefore makes it easier for me to spot the actual important changes. And if we don't align with these changes, they'll show up in the diff of every future update.

On a side note, I am personally in favor of moving things to a closer .gitignore if possible, since it reduces the chance that a rename of a parent directory will break something.

android/app/src/main/AndroidManifest.xml
Abestanis: Yeah, this is now set in the gradle file.

Abestanis:
> I am not sure if we even want to do this, the whole app design is currenly more about material 2 (and it seems OK honestly)

(GitHub doesn't let me answer directly to your comment, so I'm doing it here)

True, as long as they don't deprecate material 2 it should be fine.

nt4f04uNd (approved):

PR #131, opened 2024-07-10
Fix some warnings and replace deprecated items

Abestanis:
*The MR is build on top of #130.*

This fixes some warnings that have popped up in the new Flutter version.

There are some problematic areas:
* `textScaleFactor` is deprecated in favour of `textScaler` to support non-linear scaling:
  Except for a few instances I did not do anything with this warning, because it is not a simple replacement. We rely on the the textscale factor in some places for layout. Replacing the factor with the text-scaler will require us to provide the font size of the text that we are trying to account for. I left this for a future merge request.
* `ColorScheme` `background` and `onBackground` are deprecated in favour of `surface` and `onSurface`. Unfortunately we were already using different colors for `surface` and `onSurface`, but `background` was the same color as `secondaryContainer`, so I added `onSecondaryContainer` with the old `onBackground` color and used that throughout the app.


lib/logic/player/music_player.dart
```diff
@@ -292,7 +292,7 @@ class AudioHandler extends BaseAudioHandler with SeekHandler, WidgetsBindingObse
   Future<void> playFromUri(Uri uri, [Map<String, dynamic>? extras]) {
     // TODO: implement playFromUri
     throw UnimplementedError();
-    return super.playFromUri(uri, extras);
+    // return super.playFromUri(uri, extras);
```
nt4f04uNd, line 295: Need to remove (here and other places)

nt4f04uNd:
> ColorScheme background and onBackground are deprecated in favour of surface and onSurface. Unfortunately we were already using different colors for surface and onSurface, but background was the same color as secondaryContainer, so I added onSecondaryContainer with the old onBackground color and used that throughout the app.


I'd prefer if we actually fixed this, maybe by putting some the colors into a custom theme. OK to do it in a separate PR.

nt4f04uNd (approved):

lib/logic/player/music_player.dart
Abestanis: Done 👍

Abestanis:
> I'd prefer if we actually fixed this, maybe by putting some the colors into a custom theme. OK to do it in a separate PR.

Agreed, created #159 to track it.

Issue #98, opened 2023-05-05
Selection - playlists "deselect all" action doesn't work


nt4f04uNd:
### Steps to reproduce

1. Go to playlists tab
2. Select all
3. Tap the button once again

### What did you expect?

All deselected, therefore selection closes

### What happened instead?

Nothing happens

### Sweyer version

1.0.9

nt4f04uNd:
The problem is with any type of content, which you initially choose for selection (except for songs)

Issue #137, opened 2024-07-13
Refactor runtimeType -> contentType


nt4f04uNd:
### Suggestion

This can cause problems like this https://github.com/nt4f04uNd/sweyer/issues/98

Whe should refactore all usages of runtimeType to use contentType

PR #138, opened 2024-07-13
Fix deselect all

nt4f04uNd:
Fixes https://github.com/nt4f04uNd/sweyer/issues/98

Problem was on this line
https://github.com/nt4f04uNd/sweyer/blob/fix/98/playlist-deselect-all/lib/routes/home_route/search_route.dart#L619

Also, here this code couldn't pick up the proper `runtimeType`

I refactored both places for double-bulletproof + other place where we were checking runtimeType, instead of `contentType`

Opened an issue to remove all usages of runtimeType
https://github.com/nt4f04uNd/sweyer/issues/137

### DIFF
--- a/lib/widgets/selection.dart
+++ b/lib/widgets/selection.dart
@@ -2342,7 +2342,7 @@ class SelectAllSelectionAction<T extends Content> extends StatelessWidget {
       final first = all.first;
       assert(all.every((el) => el.runtimeType == first.runtimeType));
 
-      final selectedOfThisType = controller.data.where((el) => el.runtimeType == entryFactory(first, 0).runtimeType);
+      final selectedOfThisType = controller.data.where((el) => el.data.type == entryFactory(first, 0).data.type);
       bool allSelected = true;
       for (int i = 0; i < all.length; i++) {
         if (!selectedOfThisType.contains(entryFactory(all[i], i))) {
--- a/lib/logic/models/selection_entry.dart
+++ b/lib/logic/models/selection_entry.dart
@@ -27,13 +27,23 @@ class SelectionEntry<T extends Content> {
           origin: selectionRouteOf(context) && song.origin is DuplicatingSongOriginMixin ? song.origin : null,
         ) as SelectionEntry<T>;
       case ContentType.album:
-      case ContentType.playlist:
-      case ContentType.artist:
-        return SelectionEntry<T>(
+        return SelectionEntry<Album>(
           index: index,
-          data: content,
+          data: content as Album,
           origin: null,
-        );
+        ) as SelectionEntry<T>;
+      case ContentType.playlist:
+        return SelectionEntry<Playlist>(
+          index: index,
+          data: content as Playlist,
+          origin: null,
+        ) as SelectionEntry<T>;
+      case ContentType.artist:
+        return SelectionEntry<Artist>(
+          index: index,
+          data: content as Artist,
+          origin: null,
+        ) as SelectionEntry<T>;
     }
   }
 
--- a/lib/logic/player/content.dart
+++ b/lib/logic/player/content.dart
@@ -988,16 +988,19 @@ class ContentUtils {
     final List<SelectionEntry<Playlist>> playlists = [];
     final List<SelectionEntry<Artist>> artists = [];
     for (final entry in data) {
-      if (entry is SelectionEntry<Song>) {
-        songs.add(entry);
-      } else if (entry is SelectionEntry<Album>) {
-        albums.add(entry);
-      } else if (entry is SelectionEntry<Playlist>) {
-        playlists.add(entry);
-      } else if (entry is SelectionEntry<Artist>) {
-        artists.add(entry);
-      } else {
-        throw UnimplementedError();
+      switch (entry.data.type) {
+        case ContentType.song:
+          songs.add(entry as SelectionEntry<Song>);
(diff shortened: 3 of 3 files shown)


lib/widgets/selection.dart
```diff
@@ -2342,7 +2342,7 @@ class SelectAllSelectionAction<T extends Content> extends StatelessWidget {
       final first = all.first;
       assert(all.every((el) => el.runtimeType == first.runtimeType));
 
-      final selectedOfThisType = controller.data.where((el) => el.runtimeType == entryFactory(first, 0).runtimeType);
+      final selectedOfThisType = controller.data.where((el) => el.data.type == entryFactory(first, 0).data.type);
```
nt4f04uNd, line 2345: This solves the problem, but the same does the change in `selection_entry.dart`

Abestanis: Oh, I see, the first entry had a `runtimeType`  of  `SelectionEntry<T>`, because we are returning that in `SelectionEntry.fromContent`. But all other entries had a `runtimeType`  of `SelectionEntry<Content>`, because `_selectionPack` didn't cast the `entry`.

This is indeed very problematic, we should avoid `runtimeType` as much as possible.

lib/widgets/selection.dart
```diff
@@ -2342,7 +2342,7 @@ class SelectAllSelectionAction<T extends Content> extends StatelessWidget {
       final first = all.first;
       assert(all.every((el) => el.runtimeType == first.runtimeType));
```
Abestanis, line 2343:
```suggestion
      assert(all.every((el) => el.type == first.type));
```
Let's not use the `runtimeType` here while we are a

Abestanis (changes requested):

Abestanis:
Also, should we add a test for this to avoid regressing this in the future?

Abestanis:
I am trying to write a test for this, but I can't get the last `pumpAndSettle` to wait for the animation when deselecting everything. Maybe you have an idea what's wrong here.

```dart
group('common selection actions on all tabs', () {
    final album0 = albumWith(id: 0);
    final album1 = albumWith(id: 1);
    final album2 = albumWith(id: 2);
    final artist0 = artistWith(id: 0);
    final artist1 = artistWith(id: 1);
    final artist2 = artistWith(id: 2);
    final song0 = songWith(id: 0, title: 'Song 0', albumId: album0.id, artistId: artist0.id);
    final song1 = songWith(id: 1, title: 'Song 1', albumId: album1.id, artistId: artist1.id);
    final song2 = songWith(id: 2, title: 'Song 2', albumId: album2.id, artistId: artist2.id);
    final playlist0 = playlistWith(id: 0, songIds: [song0.id]);
    final playlist1 = playlistWith(id: 1, songIds: [song1.id]);
    final playlist2 = playlistWith(id: 2, songIds: [song2.id]);

    for (var (tabName, contentType) in [
      ('tracks', ContentType.song),
      ('album', ContentType.album),
      ('playlists', ContentType.playlist),
      ('artists', ContentType.artist),
    ]) {
      testWidgets('allows selecting all and deselecting all on the $tabName tab', (WidgetTester tester) async {
        await setUpAppTest(() {
          FakeSweyerPluginPlatform.instance.songs = [song0, song1, song2];
          FakeSweyerPluginPlatform.instance.albums = [album0, album1, album2];
          FakeSweyerPluginPlatform.instance.playlists = [playlist0, playlist1, playlist2];
          FakeSweyerPluginPlatform.instance.artists = [artist0, artist1, artist2];
        });
        await tester.runAppTest(() async {
          HomeRouter.instance.goto(HomeRoutes.tabs);
          await tester.pumpAndSettle();

          // Select tab
          if (contentType != ContentType.song) {
            await tester.tap(find.byIcon(contentType.icon).first);
            await tester.pumpAndSettle();
          }

          // Select first content
          await tester.longPress(find.byType(ContentTile).first);
          await tester.pumpAndSettle();

          var selectionController = ContentControl.instance.selectionNotifier.value!;
          expect(selectionController.inSelection, true);
          expect(selectionController.data.length, 1);
          expect(find.descendant(of: find.byType(SelectionCounter), matching: find.text('1')).hitTestable(),
              findsOneWidget);

          // Select all
          await tester.tap(find.byType(SelectAllSelectionAction).last);
          await tester.pumpAndSettle();

          int numElements = ContentControl.instance.getContent(contentType).length;
          expect(selectionController.inSelection, true);
          expect(selectionController.data.length, numElements);
          expect(find.descendant(of: find.byType(SelectionCounter), matching: find.text('$numElements')).hitTestable(),
              findsOneWidget);

          // Deselect all
          await tester.tap(find.byType(SelectAllSelectionAction).last);
          await tester.pumpAndSettle();

          expect(selectionController.inSelection, false);
          expect(selectionController.data.length, 0);
          expect(selectionController.animationController.isCompleted, true); // This fails
          expect(selectionController.animation.isCompleted, true); // This fails

          expect(find.byType(SelectAllSelectionAction), findsNothing); // This fails
          expect(find.byType(SelectionCounter), findsNothing); // This fails
        });
      });
    }
  });
```

Abestanis (approved):
A have created #142 that adds the tests for this, so we can merge this. 👍

nt4f04uNd:
> A have created #142 that adds the tests for this, so we can merge this. 👍

Thanks!

PR #140, opened 2024-07-13
Adjust workflow runs 2

nt4f04uNd:
I forgot that we have an old-style naming of the branch :-)


Abestanis (approved):
Ups, me too. We should just switch `master` to `main`.

Issue #127, opened 2024-06-29
Increase robustness when scanning for content


Abestanis:
### Suggestion

To prevent issues like #122, we should:
- [ ] Be more robust and don't fail the entire content search because one item fails. We should catch exceptions on an item level and ignore them.
- [ ] Catch errors in the content search and show an error UI instead of hanging forever.
- [ ] Handle more fields being null. It looks like all columns in the Android media database can based on the [create statement here](https://cs.android.com/android/platform/superproject/main/+/main:packages/providers/MediaProvider/src/com/android/providers/media/DatabaseHelper.java;l=1108-1144) and the fact that the album database is just [a view into the `files` database](https://cs.android.com/android/platform/superproject/main/+/main:packages/providers/MediaProvider/src/com/android/providers/media/DatabaseHelper.java;l=1606-1622), so we should be more resilient against nulls.

Abestanis:
Also, off topic, shouldn't we undraft https://github.com/nt4f04uNd/sweyer/releases/tag/untagged-88d5c1dd963c02bce466 ?

nt4f04uNd:
> Also, off topic, shouldn't we undraft https://github.com/nt4f04uNd/sweyer/releases/tag/untagged-88d5c1dd963c02bce466 ?

True

nt4f04uNd:
Your suggestions seem reasonable

nt4f04uNd:
> It looks like all columns in the Android media database can be null

This is crazy to me, by the way

Abestanis:
Yes, we already improved the situation immensely by only ignoring individual content items when they are missing any data that we require, but ideally we would handle all missing non-essential data.

For example, currently a song without a title will be ignored, but we should accept it and just show a "<Unnamed song>" or something like that. I'll go through and create sub-issues for all the attributes where we should accept `null` values.

PR #148, opened 2024-07-21
If the media store favourite state is unknown, treat it as no favourite

Abestanis:
While adding tests to work towards #127, I noticed that we would crash if `isFavoriteInMediaStore` was `null`. We haven't seen this on Crashlytics yet, but this can definitely happen on older (pre `R`, API level `30`) devices.

### DIFF
--- a/lib/logic/player/favorites.dart
+++ b/lib/logic/player/favorites.dart
@@ -54,7 +54,7 @@ class FavoritesControl with Control {
     final Set<int> favoriteSet = {};
     if (_useMediaStoreFavorites(contentType)) {
       for (final song in ContentControl.instance.state.allSongs.songs) {
-        if (song.isFavoriteInMediaStore!) {
+        if (song.isFavoriteInMediaStore ?? false) {
           favoriteSet.add(song.sourceId);
         }
       }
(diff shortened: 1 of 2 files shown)


lib/logic/player/favorites.dart
```diff
@@ -54,7 +54,7 @@ class FavoritesControl with Control {
     final Set<int> favoriteSet = {};
     if (_useMediaStoreFavorites(contentType)) {
       for (final song in ContentControl.instance.state.allSongs.songs) {
-        if (song.isFavoriteInMediaStore!) {
```
nt4f04uNd, line 57: This code must not be called on API <30

Currently it's filtered out by [this condition](https://github.com/nt4f04uNd/sweyer/blob/master/lib/logic/player/favorites.dart#L27), that's why there's a non-nullable assertion `!` there

If it would be called, I would like for this code to rather crash, than silently be executed

Abestanis: `isFavoriteInMediaStore` is just an attribute that we get from the Android media database and we've seen that they can be null, also on API >=30. I really don't think we should allow our app to crash just because that value can be null, especially here during initialization which will prevent the whole app from loading up.

An alternative approach would be to change our model and make it not-nullable, and fall back to `false` if it is null in the media database.

nt4f04uNd: I'd prefer for it to have a default value then, yeah

nt4f04uNd: Actually, I see that on the Kotlin side on R and above we send `getInt() == 1`, so it will never be null in Dart

So technically, it can be null, but practically it currently will not

Still, putting a default value would be OK

Abestanis: > Actually, I see that on the Kotlin side on R and above we send getInt() == 1, so it will never be null in Dart

Ah yes, indeed. But I agree, let's use a default value and make it not nullable.

nt4f04uNd (approved):

PR #145, opened 2024-07-14
Don't error when a playlist has no modification date

Abestanis:
Use the creation date if no modification date is present.

[...]

PR #149, opened 2024-07-21
Improve content scanning

Abestanis:
*This includes #145 and is based on top of #148.*

This makes content scanning more robust and prevents failing the entire content search because one item fails by catching exceptions on an item level and reporting them, which implements the first checkbox on #127.

I think it is valuable to report these as non-fatal errors, so we can see if we are too strict and need to make a content attribute nullable.

### DIFF
--- a/test/logic/models/content_test.dart
+++ b/test/logic/models/content_test.dart
@@ -3,7 +3,7 @@ import 'package:collection/collection.dart';
 import '../../observer/observer.dart';
 import '../../test.dart';
 
-extension _TestMap on Map<String, dynamic> {
+extension on Map<String, dynamic> {
   Map<String, dynamic> copyWith(Map<String, dynamic> overwrites) {
     return Map.of(this)..addAll(overwrites);
   }
@@ -63,7 +63,7 @@ extension _TestMap on Map<String, dynamic> {
   }
 }
 
-extension _TestListMap on List<Map<String, dynamic>> {
+extension on List<Map<String, dynamic>> {
   List<Map<String, dynamic>> assignUniqueIds({required int startId}) {
     for (final element in this) {
       if (element['id'] is int) {
--- a/test/logic/models/content_test.dart
+++ b/test/logic/models/content_test.dart
@@ -164,18 +164,12 @@ void main() {
         (validPlaylist.copyWith({'id': 4}).copyWithout(['filesystemPath']), playlistWith(id: 4, fileSystemPath: null)),
         (validPlaylist.copyWith({'id': 5, 'dateAdded': -1}), playlistWith(id: 5, dateAdded: -1)),
         (validPlaylist.copyWith({'id': 6, 'dateModified': -1}), playlistWith(id: 6, dateModified: -1)),
-        (validPlaylist.copyWith({'id': 7, 'name': ''}), playlistWith(id: 7, name: '')),
-        (validPlaylist.copyWith({'id': 8, 'songIds': []}), playlistWith(id: 8, songIds: [])),
-        (validPlaylist.copyWith({'id': 9, 'songIds': null}), playlistWith(id: 9, songIds: [])),
-        (validPlaylist.copyWith({'id': 10}).copyWithout(['songIds']), playlistWith(id: 10, songIds: [])),
-        (
-          validPlaylist.copyWith({'id': 11, 'dateAdded': 42, 'dateModified': null}),
-          playlistWith(id: 11, dateAdded: 42, dateModified: 42)
-        ),
-        (
-          validPlaylist.copyWith({'id': 12, 'dateAdded': 42}).copyWithout(['dateModified']),
-          playlistWith(id: 12, dateAdded: 42, dateModified: 42)
-        ),
+        (validPlaylist.copyWith({'id': 7, 'dateModified': null}), playlistWith(id: 7, dateModified: null)),
+        (validPlaylist.copyWith({'id': 8}).copyWithout(['dateModified']), playlistWith(id: 8, dateModified: null)),
+        (validPlaylist.copyWith({'id': 9, 'name': ''}), playlistWith(id: 9, name: '')),
+        (validPlaylist.copyWith({'id': 10, 'songIds': []}), playlistWith(id: 10, songIds: [])),
+        (validPlaylist.copyWith({'id': 11, 'songIds': null}), playlistWith(id: 11, songIds: [])),
+        (validPlaylist.copyWith({'id': 12}).copyWithout(['songIds']), playlistWith(id: 12, songIds: [])),
         (
           validPlaylist.copyWith({
             'id': 13,
@@ -189,7 +183,6 @@ void main() {
         ...validPlaylist.copyWithout(propertiesThatCanBeMissing).subSets(),
         ...validPlaylist.withWrongTypes().whereNot(
             (map) => propertiesThatCanBeMissing.any((property) => map[property] == null) || map['songIds'] == []),
-        validPlaylist.copyWith({'dateAdded': null, 'dateModified': null}),
       ].assignUniqueIds(startId: validPlaylists.last.$2.id + 1);
       late CrashlyticsObserver crashlyticsObserver;
       await setUpAppTest(() {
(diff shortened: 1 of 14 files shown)


test/logic/models/content_test.dart
```diff
@@ -0,0 +1,297 @@
...
+        (validAlbum.copyWith({'id': 15}).copyWithout(['lastYear']), albumWith(id: 15, lastYear: null)),
+        (validAlbum.copyWith({'id': 16, 'numberOfSongs': -1}), albumWith(id: 16, numberOfSongs: -1)),
+        (validAlbum.copyWith({'id': 17, 'numberOfSongs': null}), albumWith(id: 17, numberOfSongs: 0)),
+        (validAlbum.copyWith({'id': 18}).copyWithout(['numberOfSongs']), albumWith(id: 18, numberOfSongs: 0)),
+        (
+          validAlbum.copyWith({'id': 19, 'lastYear': 2000, 'firstYear': 3000}), // Last year before first year
+          albumWith(id: 19, lastYear: 2000, firstYear: 3000),
+        ),
+      ];
+      final propertiesThatCanBeMissing = ['albumArt', 'artistId', 'firstYear', 'lastYear', 'numberOfSongs'];
```
nt4f04uNd, line 105: If we add a new property, this will likely not be updated

test/logic/models/content_test.dart
```diff
@@ -0,0 +1,297 @@
+import 'package:collection/collection.dart';
+
+import '../../observer/observer.dart';
+import '../../test.dart';
+
+extension TestMap on Map<String, dynamic> {
```
nt4f04uNd, line 6: I'd prefer to not make it public

test/logic/models/content_test.dart
```diff
@@ -0,0 +1,297 @@
...
+        'filesystemPath',
+        'isFavoriteInMediaStore',
+        'generationAdded',
+        'generationModified',
+      ];
+      final invalidSongs = [
+        ...validSong.copyWithout(propertiesThatCanBeMissing).subSets(),
+        ...validSong
+            .withWrongTypes()
+            .whereNot((map) => propertiesThatCanBeMissing.any((property) => map[property] == null))
```
nt4f04uNd, line 280: What does it do?

test/logic/models/content_test.dart
```diff
@@ -0,0 +1,297 @@
...
+import '../../observer/observer.dart';
+import '../../test.dart';
+
+extension TestMap on Map<String, dynamic> {
+  Map<String, dynamic> copyWith(Map<String, dynamic> overwrites) {
+    return Map.of(this)..addAll(overwrites);
+  }
+
+  Map<String, dynamic> copyWithout(List<String> keysToRemove) {
+    var map = Map.of(this);
```
nt4f04uNd, line 12: final

test/logic/models/content_test.dart
```diff
@@ -0,0 +1,297 @@
...
+        (validAlbum.copyWith({'id': 17, 'numberOfSongs': null}), albumWith(id: 17, numberOfSongs: 0)),
+        (validAlbum.copyWith({'id': 18}).copyWithout(['numberOfSongs']), albumWith(id: 18, numberOfSongs: 0)),
+        (
+          validAlbum.copyWith({'id': 19, 'lastYear': 2000, 'firstYear': 3000}), // Last year before first year
+          albumWith(id: 19, lastYear: 2000, firstYear: 3000),
+        ),
+      ];
+      final propertiesThatCanBeMissing = ['albumArt', 'artistId', 'firstYear', 'lastYear', 'numberOfSongs'];
+      final invalidAlbums = [
+        ...validAlbum.copyWithout(propertiesThatCanBeMissing).subSets(),
```
nt4f04uNd, line 107: What is the point of generating a model, then removing by hand all of its properties?

nt4f04uNd (commented):
General comment: the solution is fine, but tests are hard to read. If we could make it a bit more simple, or more readable, that would be great

test/logic/models/content_test.dart
nt4f04uNd: Generally it's hard to understand what is the output of these function calls, how they work, and why we use them

test/logic/models/content_test.dart
Abestanis: The way I see it, that's not going to be a problem. The tests verify the current behaviour, any change will (and should) break them. If we add a nullable property, the `subSets` function below will generate albums maps where that property is null/missing, so the test will fail when we expect parsing of these to fail but they get accepted.

test/logic/models/content_test.dart
Abestanis: 👍 Done.

test/logic/models/content_test.dart
```diff
@@ -3,7 +3,7 @@ import 'package:collection/collection.dart';
 import '../../observer/observer.dart';
 import '../../test.dart';
 
-extension TestMap on Map<String, dynamic> {
+extension _TestMap on Map<String, dynamic> {
```
nt4f04uNd, line 6: you can remove the extension name, this will make it private

Abestanis: Ah, even better, thanks. I didn't know that was possible.

test/logic/models/content_test.dart
Abestanis: `subSets` generates all possible variations of the map where one or more elements are missing to generate a list of invalid maps that can't be parsed into a valid album. The `copyWithout` removes the elements of the map that are allowed to be missing and would therefore successfully parse.

The reason to not hard code the list of properties is that if we add a new property to `Album` it will be automatically included here.

nt4f04uNd: > The copyWithout removes the elements of the map that are allowed to be missing and would therefore successfully parse.

But if it removes those fields from map, subSets gets the map with those removed fields. How would it know it needs to generate some variations of those non-existent fields, am I missing something here?

test/logic/models/content_test.dart
Abestanis: > Generally it's hard to understand what is the output of these function calls, how they work, and why we use them

I agree.

The point of this is to automatically create all possible bad variations of inputs to the `parse` function.
There is two types of problems that this covers:
* Missing data: The `subSets` function gives us variations of the `validSong` map (without the values that are allowed to be null) with one or more missing elements. We generate all possible subsets instead of just checking if it's ok if individual elements are missing because sometimes we fall back to another element in that case (like `Song.dateModified` falling back to `Song.dateAdded`).
* Invalid data types: The `withWrongTypes` gives us variations of the `validSong` map where each element of the map is a different type from the expected one. For example, the `title` property has a type `String`, so `withWrongTypes` will generate one variation where the `title` is `null`, one where the `title` is `true`, one where it's `0` and so on. We only generate variations where one of the elements is changed and don't bother with variations where multiple element types have changed because there would be a large amount of variations and because I don't believe it is very relevant.

I agree that this is not very easy to parse from the code, but I'm not quite sure how to improve the situation other than adding comments and maybe changing the name of `withWrongTypes` and `subSets`. Do you have any recomendations?

test/logic/models/content_test.dart
Abestanis: > How would it know it needs to generate some variations of those non-existent fields, am I missing something here?

The idea is for it to not generate any variations with the removed fields. We want to generate invalid maps here, but since these field are nullable it's valid for them to be missing.

test/logic/models/content_test.dart
Abestanis: I've added some helper extension methods to improve the clarity of the tests.

Abestanis: I think splitting the tests into the valid and invalid cases improves clarity a lot.

Abestanis:
@nt4f04uNd I believe I have addressed all of your previous comments, let me know if you are ok with this change now. From my side this is ready.

nt4f04uNd (approved):

Abestanis:
A some small adjustments to the tests were necessary to rebase on top of #181, but no actual changes on the app code were necessary, so I assume your approval still holds.

PR #152, opened 2024-08-25
Add tests for sorting behaviour and make album year optional

Abestanis:
This is part of my quest to make the app more robust by accepting more `null` values in the content (aka the third point of #127). If we want to make certain content fields nullable we will have to change how sorting works, so I wanted to add some tests for the existing sorting behaviour.

When adding the tests I noticed that the album `year` getter falls back to the year of the current system time. This does not make sense to me, because that means that sorting depends on the current year (or if the device is not syncing time online, whatever year it thinks it is).

I also changed the `*SortFeature`s into regular enums. Whatever prevented them from being regular enums in the past seems to be part of the dart language now, so it seems to be better to use actual enums.


lib/localization/localization.dart
```diff
@@ -65,8 +65,6 @@ extension AppLocalizationsExtension on AppLocalizations {
             return artist;
           case SongSortFeature.album:
             return album;
-          default:
```
Abestanis, line 68: Since the `*SortFeature`s are now regular enums, we can exhaustively match on them. This means that if we add a new `SongSortFeature` we won't forget to add another case here because we'll be notified at compiletime, instead of runtime.

lib/logic/models/album.dart
```diff
@@ -37,20 +37,15 @@ class Album extends PersistentQueue implements PlatformAlbum {
   bool get playable => true;
 
   /// Gets album normalized year.
-  int get year {
-    return lastYear == null || lastYear! < 1000 ? clock.now().year : lastYear!;
+  int? get year {
+    return lastYear ?? firstYear;
```
Abestanis, line 41: I don't think using `clock.now().year` makes sense.

We could potentially add some logic to convert `0 <= year < 50` to `2000 + year` and `50 <= year < 100` to `1900 + year`, but I'm not sure if that's a good long term solution to have that.

And I think falling back to `firstYear` makes sense either way.

lib/logic/models/album.dart
```diff
@@ -37,20 +37,15 @@ class Album extends PersistentQueue implements PlatformAlbum {
...
   }
 
   /// Returns string in format `album name • year`.
   String get nameDotYear {
     return ContentUtils.appendYearWithDot(album, year);
   }
 
-  /// Returns string in format `Album • year`.
-  String albumDotName(AppLocalizations l10n) {
```
Abestanis, line 50: This was just unused.

Abestanis:
+1.05% test coverage

Nice, that increase in test coverage is what I was hoping to see. 😄

lib/logic/models/sort.dart
```diff
@@ -292,13 +271,8 @@ class AlbumSort extends Sort<Album> {
...
   }
 
-  @override
-  AlbumSort get withDefaultOrder {
-    return copyWith(orderAscending: feature.defaultOrderAscending);
-  }
-
   int _fallbackYear(Album a, Album b) {
-    return a.year.compareTo(b.year);
+    return (a.year ?? 0).compareTo((b.year ?? 0)); // TODO: Decide how to sort null values
```
Abestanis, line 275: I left this as a todo for a later PR, but we should discuss how we sort null values. I would say they should always be at the end, no matter what (ignoring ascending/descending). What's your opinion on this?

<details><summary>I have some nice extension methods that will make it easy that I'll add in another PR.</summary>

```dart
extension NullableCompare<T> on T? {
  int compareTo(T? other) {
    final self = this;
    if (self != null) {
      self.compareToNullable(other);
    }
    if (other != null) {
      return -other.compareToNullable(self);
    }
    return 0;
  }
}

extension CompareNullable<T> on dynamic {
  int compareToNullable(T? other) {
    if (other != null) {
      return compareTo(other);
    }
    return -1;
  }
}
```
</details>

lib/logic/player/content.dart
```diff
@@ -822,18 +822,8 @@ class ContentUtils {
...
-    var result = list.first;
-    for (int i = 1; i < list.length; i++) {
-      final string = list[i].toString();
-      if (string.isNotEmpty) {
-        result += ' $dot $string';
-      }
-    }
-    return result;
+  static String joinDot(List<Object?> list) {
+    return list.where((item) => item != null && item.toString().isNotEmpty).join(' $dot ');
```
Abestanis, line 826: This needed to change to handle `null` values.

lib/logic/models/sort.dart
nt4f04uNd: SGTM

test/logic/models/sort_test.dart
```diff
@@ -0,0 +1,201 @@
...
+      final sortOrder = entry.key;
+      final featureTestCases = entry.value;
+      for (final entry in featureTestCases.entries) {
+        final feature = entry.key;
+        final expectedContentList = entry.value;
+        test('Sorts songs by ${feature.name} ${sortOrder.name}', () async {
+          expect(
+            songs.toList()
+              ..sort(SongSort(feature: feature, orderAscending: sortOrder == SortOrder.ascending).comparator),
+            expectedContentList.map((i) => songs[i]).toList(),
```
nt4f04uNd, line 53: Those test cases would be much more readable, if instead of IDs we would use the objects directly

You could so see the initial list vs expected list easily

Right now it's very hard to understand them

nt4f04uNd: Image one of those tests start to fail

I imagine determining what went wrong will be a real pain, but with `expect(list, expectedList)` it will show you the diff right in the error message

lib/logic/models/sort.dart
```diff
@@ -1,15 +1,19 @@
...
+
+  /// Descending sort order ("higher" values first, "lower" values last).
+  descending,
+}
 
+/// Interface for other sort feature enums.
+abstract class SortFeature<T extends Content> {
   /// Returns sort feature values for a given content.
   static List<SortFeature<T>> getValuesForContent<T extends Content>(ContentType<T> contentType) {
-    // TODO: Remove ContentType cast, see https://github.com/dart-lang/language/issues/2315
```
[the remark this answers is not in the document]
Abestanis: Because https://github.com/dart-lang/language/issues/2315 has gotten no traction in the last two years, and since with the new Flutter update there is no longer a lint against this, I think it's acceptable to leave the cast. The original intent of the todo was to be able to get rid of the lint ignore below.

lib/logic/models/sort.dart
```diff
@@ -22,158 +26,147 @@ abstract class SortFeature<T extends Content> extends Enum<String> {
...
-    required SongSortFeature feature,
-    bool orderAscending = true,
-  }) : super(feature: feature, orderAscending: orderAscending);
+    required SongSortFeature super.feature,
+    super.orderAscending = true,
+  });
   SongSort.defaultOrder(feature) : super.defaultOrder(feature);
 
   factory SongSort.fromMap(Map map) => SongSort(
-        feature: EnumToString.fromString(
```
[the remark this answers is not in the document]
Abestanis: Yes, I have a follow-up PR in the works that deals with the null-sorting issue where I get rid of it.

test/logic/models/sort_test.dart
Abestanis: I'm not sure I understand what you mean. Here we are expecting on `List<Song>`, not `List<int>`, so a failure looks like this:
```
Expected: [
            Song:Song(16),
            Song:Song(14),
            Song:Song(13),
            Song:Song(15),
            Song:Song(8),
            Song:Song(6),
            Song:Song(5),
            Song:Song(7),
            Song:Song(0),
            Song:Song(1),
            Song:Song(2),
            Song:Song(3),
            Song:Song(4),
            Song:Song(9),
            Song:Song(10),
            Song:Song(12),
            Song:Song(11)
          ]
  Actual: [
            Song:Song(16),
            Song:Song(14),
            Song:Song(13),
            Song:Song(15),
            Song:Song(8),
            Song:Song(6),
            Song:Song(5),
            Song:Song(7),
            Song:Song(0),
            Song:Song(1),
            Song:Song(2),
            Song:Song(3),
            Song:Song(4),
            Song:Song(9),
            Song:Song(10),
            Song:Song(11),
            Song:Song(12)
          ]
   Which: at location [15] is Song:<Song(11)> instead of Song:<Song(12)>
```

Maybe you are saying that testcases should be a `Map<SortOrder, Map<SongSortFeature, List<Song>>>`, but I'm not really convinced that either creating a new song with `songWith` to match one we created in the `setUp` or referencing the song created in `setUp` will be any more readable.

lib/logic/models/sort.dart
```diff
@@ -22,158 +26,147 @@ abstract class SortFeature<T extends Content> extends Enum<String> {
...
+  String get id => this.name;
+
+  @override
+  final SortOrder defaultSortingOrder;
 }
 
 abstract class Sort<T extends Content> extends Equatable {
   const Sort({
     required this.feature,
     required this.orderAscending,
```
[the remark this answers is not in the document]
Abestanis: Yeah, I guess it's clearer with the enum.

test/logic/models/sort_test.dart
nt4f04uNd: About the error message - yeah, it looks good. I thought it would look different.

> Maybe you are saying that testcases should be a Map<SortOrder, Map<SongSortFeature, List<Song>>>, but I'm not really convinced that either creating a new song with songWith to match one we created in the setUp or referencing the song created in setUp will be any more readable.

Yes, that's what I meant. I would like for me to be able to humanly read those test cases and verify that the expectations have the songs actually in the order that I would expect (by seeing the song parameters)

Thinking about it more, maybe we should even compare `song.toMap`s, because now it only compares songs by IDs - it would also make the error message more detailed, and disallow any discrepancies that might happen if we implement what I suggested above

What do you think?

lib/logic/models/sort.dart
```diff
@@ -1,15 +1,19 @@
...
+enum SortOrder {
+  /// Ascending sort order ("lower" values first, "higher" values last).
+  ascending,
+
+  /// Descending sort order ("higher" values first, "lower" values last).
+  descending,
+}
 
+/// Interface for other sort feature enums.
+abstract class SortFeature<T extends Content> {
```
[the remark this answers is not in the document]
Abestanis: Ah, the CI reminded me why I didn't do it (yet). The dart version we use on `master` doesn't support that combination yet. We'll have to wait for #130 to be merged.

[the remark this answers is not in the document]
Abestanis: And then a moment later I see that you approved that PR, so we can use `abstract interface class`. Sorry for the noise.

lib/logic/models/sort.dart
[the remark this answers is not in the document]
Abestanis: Done, I had to add a fallback when deserializing the sort configuration from a map to not break any existing installations.

test/logic/models/sort_test.dart
Abestanis: I think you are right. My fear was that by creating these huge lists of song constructors it would be just overwhelming and less readable, and while I still think it's a huge wall of text, I agree that it helps to understand how we are actually sorting.

<details><summary>Example of a album test failure</summary>

```
Expected: [
            {
              'id': 10,
              'album': '',
              'albumArt': 'albumArt_albumArt_albumArt',
              'artist': '',
              'artistId': 0,
              'firstYear': 0,
              'lastYear': -1,
              'numberOfSongs': 0
            },
            {
              'id': 13,
              'album': '',
              'albumArt': 'albumArt_albumArt_albumArt',
              'artist': '',
              'artistId': 0,
              'firstYear': -1,
              'lastYear': null,
              'numberOfSongs': 0
            },
            {
              'id': 0,
              'album': '',
              'albumArt': 'albumArt_albumArt_albumArt',
              'artist': '',
              'artistId': 0,
              'firstYear': 0,
              'lastYear': 0,
              'numberOfSongs': 0
            },
            {
              'id': 5,
              'album': '',
              'albumArt': 'albumArt_albumArt_albumArt',
              'artist': 'A',
              'artistId': 0,
              'firstYear': 0,
              'lastYear': 0,
              'numberOfSongs': 0
            },
            {
              'id': 6,
              'album': '',
              'albumArt': 'albumArt_albumArt_albumArt',
              'artist': 'AA',
              'artistId': 0,
              'firstYear': 0,
              'lastYear': 0,
              'numberOfSongs': 0
            },
            {
              'id': 7,
              'album': '',
              'albumArt': 'albumArt_albumArt_albumArt',
              'artist': 'a',
              'artistId': 0,
              'firstYear': 0,
              'lastYear': 0,
              'numberOfSongs': 0
            },
            {
              'id': 8,
              'album': '',
              'albumArt': 'albumArt_albumArt_albumArt',
              'artist': 'z',
              'artistId': 0,
              'firstYear': 0,
              'lastYear': 0,
              'numberOfSongs': 0
            },
            {
              'id': 11,
              'album': '',
              'albumArt': 'albumArt_albumArt_albumArt',
              'artist': '',
              'artistId': 0,
              'firstYear': 0,
              'lastYear': null,
              'numberOfSongs': 0
            },
            {
              'id': 14,
              'album': '',
              'albumArt': 'albumArt_albumArt_albumArt',
              'artist': '',
              'artistId': 0,
              'firstYear': null,
              'lastYear': null,
              'numberOfSongs': 0
            },
            {
              'id': 15,
              'album': '',
              'albumArt': 'albumArt_albumArt_albumArt',
              'artist': '',
              'artistId': 0,
              'firstYear': 0,
              'lastYear': 0,
              'numberOfSongs': 1
            },
            {
              'id': 16,
              'album': '',
              'albumArt': 'albumArt_albumArt_albumArt',
              'artist': '',
              'artistId': 0,
              'firstYear': 0,
              'lastYear': 0,
              'numberOfSongs': -1
            },
            {
              'id': 9,
              'album': '',
              'albumArt': 'albumArt_albumArt_albumArt',
              'artist': '',
              'artistId': 0,
              'firstYear': 0,
              'lastYear': 1,
              'numberOfSongs': 0
            },
            {
              'id': 12,
              'album': '',
              'albumArt': 'albumArt_albumArt_albumArt',
              'artist': '',
              'artistId': 0,
              'firstYear': 1,
              'lastYear': null,
              'numberOfSongs': 0
            },
            {
              'id': 1,
              'album': 'A',
              'albumArt': 'albumArt_albumArt_albumArt',
              'artist': '',
              'artistId': 0,
              'firstYear': 0,
              'lastYear': 0,
              'numberOfSongs': 0
            },
            {
              'id': 3,
              'album': 'a',
              'albumArt': 'albumArt_albumArt_albumArt',
              'artist': '',
              'artistId': 0,
              'firstYear': 0,
              'lastYear': 0,
              'numberOfSongs': 0
            },
            {
              'id': 4,
              'album': 'z',
              'albumArt': 'albumArt_albumArt_albumArt',
              'artist': '',
              'artistId': 0,
              'firstYear': 0,
              'lastYear': 0,
              'numberOfSongs': 0
            },
            {
              'id': 2,
              'album': 'AA',
              'albumArt': 'albumArt_albumArt_albumArt',
              'artist': '',
              'artistId': 0,
              'firstYear': 0,
              'lastYear': 0,
              'numberOfSongs': 0
            }
          ]
  Actual: [
            {
              'id': 10,
              'album': '',
              'albumArt': 'albumArt_albumArt_albumArt',
              'artist': '',
              'artistId': 0,
              'firstYear': 0,
              'lastYear': -1,
              'numberOfSongs': 0
            },
            {
              'id': 13,
              'album': '',
              'albumArt': 'albumArt_albumArt_albumArt',
              'artist': '',
              'artistId': 0,
              'firstYear': -1,
              'lastYear': null,
              'numberOfSongs': 0
            },
            {
              'id': 0,
              'album': '',
              'albumArt': 'albumArt_albumArt_albumArt',
              'artist': '',
              'artistId': 0,
              'firstYear': 0,
              'lastYear': 0,
              'numberOfSongs': 0
            },
            {
              'id': 5,
              'album': '',
              'albumArt': 'albumArt_albumArt_albumArt',
              'artist': 'A',
              'artistId': 0,
              'firstYear': 0,
              'lastYear': 0,
              'numberOfSongs': 0
            },
            {
              'id': 6,
              'album': '',
              'albumArt': 'albumArt_albumArt_albumArt',
              'artist': 'AA',
              'artistId': 0,
              'firstYear': 0,
              'lastYear': 0,
              'numberOfSongs': 0
            },
            {
              'id': 7,
              'album': '',
              'albumArt': 'albumArt_albumArt_albumArt',
              'artist': 'a',
              'artistId': 0,
              'firstYear': 0,
              'lastYear': 0,
              'numberOfSongs': 0
            },
            {
              'id': 8,
              'album': '',
              'albumArt': 'albumArt_albumArt_albumArt',
              'artist': 'z',
              'artistId': 0,
              'firstYear': 0,
              'lastYear': 0,
              'numberOfSongs': 0
            },
            {
              'id': 11,
              'album': '',
              'albumArt': 'albumArt_albumArt_albumArt',
              'artist': '',
              'artistId': 0,
              'firstYear': 0,
              'lastYear': null,
              'numberOfSongs': 0
            },
            {
              'id': 14,
              'album': '',
              'albumArt': 'albumArt_albumArt_albumArt',
              'artist': '',
              'artistId': 0,
              'firstYear': null,
              'lastYear': null,
              'numberOfSongs': 0
            },
            {
              'id': 15,
              'album': '',
              'albumArt': 'albumArt_albumArt_albumArt',
              'artist': '',
              'artistId': 0,
              'firstYear': 0,
              'lastYear': 0,
              'numberOfSongs': 1
            },
            {
              'id': 16,
              'album': '',
              'albumArt': 'albumArt_albumArt_albumArt',
              'artist': '',
              'artistId': 0,
              'firstYear': 0,
              'lastYear': 0,
              'numberOfSongs': -1
            },
            {
              'id': 9,
              'album': '',
              'albumArt': 'albumArt_albumArt_albumArt',
              'artist': '',
              'artistId': 0,
              'firstYear': 0,
              'lastYear': 1,
              'numberOfSongs': 0
            },
            {
              'id': 12,
              'album': '',
              'albumArt': 'albumArt_albumArt_albumArt',
              'artist': '',
              'artistId': 0,
              'firstYear': 1,
              'lastYear': null,
              'numberOfSongs': 0
            },
            {
              'id': 1,
              'album': 'A',
              'albumArt': 'albumArt_albumArt_albumArt',
              'artist': '',
              'artistId': 0,
              'firstYear': 0,
              'lastYear': 0,
              'numberOfSongs': 0
            },
            {
              'id': 3,
              'album': 'a',
              'albumArt': 'albumArt_albumArt_albumArt',
              'artist': '',
              'artistId': 0,
              'firstYear': 0,
              'lastYear': 0,
              'numberOfSongs': 0
            },
            {
              'id': 2,
              'album': 'AA',
              'albumArt': 'albumArt_albumArt_albumArt',
              'artist': '',
              'artistId': 0,
              'firstYear': 0,
              'lastYear': 0,
              'numberOfSongs': 0
            },
            {
              'id': 4,
              'album': 'z',
              'albumArt': 'albumArt_albumArt_albumArt',
              'artist': '',
              'artistId': 0,
              'firstYear': 0,
              'lastYear': 0,
              'numberOfSongs': 0
            }
          ]
   Which: at location [15]['id'] is <2> instead of <4>
```
</details>

lib/logic/models/sort.dart
```diff
@@ -1,15 +1,25 @@
...
+/// The order of a sort operation.
+enum SortOrder {
+  /// Ascending sort order ("lower" values first, "higher" values last).
+  ascending,
+
+  /// Descending sort order ("higher" values first, "lower" values last).
+  descending;
+
+  /// The inversion of the sort order.
+  SortOrder get inverted => switch (this) {
```
nt4f04uNd, line 13: Wow Dart 3 really shines with this, very neat

test/logic/models/sort_test.dart
nt4f04uNd: Thanks, that is now so much better

nt4f04uNd (approved):

nt4f04uNd:
Also created https://github.com/nt4f04uNd/sweyer/issues/158

PR #153, opened 2024-08-31
Update all GitHub actions dependencies

Abestanis:
This updates our CI dependencies. I read through the change logs and made sure that there are no breaking changes for us.

There is one exception:

`codecov/codecov-action@v4` removes tokenless uploads of coverage information. This means that we need to make sure that we have [configured our Codecov token in the GitHub repository settings as described in the documentation](https://docs.codecov.com/docs/adding-the-codecov-token#github-actions). I don't have access to these settings so only you can do that.

Hopefully the update will fix the Codecov upload failures we've seen on recent CI runs like [here](https://github.com/nt4f04uNd/sweyer/actions/runs/10121497714/job/27992507895#step:4:33).

### DIFF
--- a/.github/workflows/deploy_google_play/action.yml
+++ b/.github/workflows/deploy_google_play/action.yml
@@ -67,7 +67,7 @@ runs:
         mv key.properties ${{ github.workspace }}/android/key.properties
 
     - name: 📖 Read app version
-      uses: NiklasLehnfeld/flutter-version-number-action@main
+      uses: NiklasLehnfeld/flutter-version-number-action@v1
       id: read-version
       with:
         file-path: pubspec.yaml
--- a/.github/workflows/deploy_google_play/action.yml
+++ b/.github/workflows/deploy_google_play/action.yml
@@ -106,21 +106,21 @@ runs:
         zip -r ${{ github.workspace }}/symbols.zip ${{ github.workspace }}/build/app/outputs/symbols/
 
     - name: 📁 Upload build artifacts (symbols)
-      uses: actions/upload-artifact@v3
+      uses: actions/upload-artifact@v4
       with:
         name: symbols
         path: |
           ${{ github.workspace }}/symbols.zip
 
     - name: 📁 Upload build artifacts (aab)
-      uses: actions/upload-artifact@v3
+      uses: actions/upload-artifact@v4
       with:
         name: sweyer@${{ steps.read-version.outputs.version-number }}.aab
         path: |
           ${{ github.workspace }}/build/app/outputs/bundle/release/sweyer@${{ steps.read-version.outputs.version-number }}.aab
 
     - name: 📁 Upload build artifacts (apk)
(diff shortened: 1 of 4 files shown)


Abestanis:
As a side note, we might want to set up dependabot to keep these up to date in the future.

Abestanis:
It looks like the Codecov update did indeed help. The coverage change probably comes from an earlier merge request that failed to upload its coverage information.

Coverage from this PR

.github/workflows/deploy_google_play/action.yml
Abestanis, line 1: I can't comment on the specific line, but it there a reason we use `NiklasLehnfeld/flutter-version-number-action@main` and not `NiklasLehnfeld/flutter-version-number-action@v1`? I saw that there was one additional commit which fixes a problem when there are multiple `version:` entries in the `pubspec.yaml`, but that doesn't affect us. Is it ok for you to switch to `v1`?

nt4f04uNd: I think it should be OK to switch

nt4f04uNd:
> It looks like the Codecov update did indeed help. The coverage change probably comes from an earlier merge request that failed to upload its coverage information.
> 
> Coverage from this PR

FYI I didn't configure anything beforehand.
I only added the token now.

Should we test somehow that it works?

nt4f04uNd:
Ahh, we should use this now. I think you should update the YAML files with this

image

nt4f04uNd:
FYI secrets must be passed explicitly to actions

You can see an example here https://github.com/nt4f04uNd/sweyer/blob/master/.github/workflows/deploy_google_play/action.yml#L30

.github/workflows/deploy_google_play/action.yml
Abestanis: Done

Abestanis:
> FYI secrets must be passed explicitly to actions

Ups, I totally missed that part, thanks for double checking.

> FYI I didn't configure anything beforehand.

Yeah, the main repository now requires a token, but forks can continue to use tokenless upload ([From the codecov actions `README`](https://github.com/codecov/codecov-action?tab=readme-ov-file#v4-release)):
> Tokenless uploading is unsupported. However, PRs made from forks to the upstream public repos will support tokenless (e.g. contributors to OS projects do not need the upstream repo's Codecov token).

> Should we test somehow that it works?

I think we can just merge and verify that the CI on the merge commit works, and iterate on it in a follow up PR if anything breaks.

Abestanis:
> FYI secrets must be passed explicitly to actions

Ah, you are saying the files included with `use` don't have access to secrets? That makes sense.

nt4f04uNd:
> As a side note, we might want to set up dependabot to keep these up to date in the future.

Can you open an issue?

Abestanis:
> Can you open an issue?

Done: https://github.com/nt4f04uNd/sweyer/issues/155

nt4f04uNd (approved):

Issue #158, opened 2024-10-14
Auto test cyrillic sorting


nt4f04uNd:
### Suggestion

https://github.com/nt4f04uNd/sweyer/pull/152 adds sorting tests

Add variations with cyrillic letters

PR #161, opened 2024-10-18
Add cyrillic letters to sort tests

nt4f04uNd:
Fixes https://github.com/nt4f04uNd/sweyer/issues/158


.gitignore
```diff
@@ -71,3 +69,6 @@ temp/
 
 # scan temporary files
 **/fastlane/test_output
+
+# FVM Version Cache
+.fvm/
```
Abestanis, line 74:
```suggestion
.fvm/

```
Let's keep the newline at the end.

test/logic/models/sort_test.dart
```diff
@@ -112,19 +172,35 @@ void main() {
...
           songWith(id: 12, dateModified: 0, dateAdded: 0, title: '', artist: 'z', album: ''),
+          songWith(id: 21, dateModified: 0, dateAdded: 0, title: '', artist: 'А (кириллица)', album: ''),
+          songWith(id: 22, dateModified: 0, dateAdded: 0, title: '', artist: 'АА (кириллица)', album: ''),
+          songWith(id: 23, dateModified: 0, dateAdded: 0, title: '', artist: 'а (кириллица)', album: ''),
+          songWith(id: 24, dateModified: 0, dateAdded: 0, title: '', artist: 'я (кириллица)', album: ''),
           songWith(id: 5, dateModified: 0, dateAdded: 0, title: 'A', artist: '', album: ''),
           songWith(id: 7, dateModified: 0, dateAdded: 0, title: 'a', artist: '', album: ''),
           songWith(id: 6, dateModified: 0, dateAdded: 0, title: 'AA', artist: '', album: ''),
           songWith(id: 8, dateModified: 0, dateAdded: 0, title: 'z', artist: '', album: ''),
+          songWith(id: 17, dateModified: 0, dateAdded: 0, title: 'А (кириллица)', artist: '', album: ''),
```
Abestanis, line 183: At first I thought that it's weird that this entry is is sorted after the `z`, but it looks like it's standard behaviour that all cyrillic letters are sorted after the ascii ones.

Abestanis (approved):
Looks good, one small nit about the newline. Thanks for adding the tests.

nt4f04uNd:
> Thanks for adding the tests.

Thanks for making them =)

Abestanis:
I've seen this CI failure before, but I can't replicate it locally. My guess is that we start the initialization in a test, and then the test ends but the while the initialization is waiting on an `await` point and we start a new test overwriting the queue controller. For now I think you can just re-run the test, but a real solution it to either find a way to cancel / await all un-awaited futures started or to not have singletons and instead pass in all controllers when creating the app.

nt4f04uNd:
Created an issue https://github.com/nt4f04uNd/sweyer/issues/162

Issue #155, opened 2024-08-31
Set up Dependabot


Abestanis:
### Suggestion

We should set up Dependabot to keep our dependencies (both GitHub actions and Flutter) up to date.

PR #163, opened 2024-10-18
Create dependabot.yml

nt4f04uNd:
Fixes https://github.com/nt4f04uNd/sweyer/issues/155

Here are other settings in GitHub UI 
image

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -15,7 +15,7 @@ updates:
     schedule:
       interval: "daily"
 
- - package-ecosystem: "github-actions"
+  - package-ecosystem: "github-actions"
     directory: "/"
     schedule:
       interval: "daily"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -14,3 +14,8 @@ updates:
     directory: "/sweyer_plugin"
     schedule:
       interval: "daily"
+
+ - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "daily"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,16 @@
...
+updates:
+  - package-ecosystem: "pub"
+    directory: "/"
+    schedule:
+      interval: "daily"
+
+  - package-ecosystem: "pub"
+    directory: "/sweyer_plugin"
+    schedule:
+      interval: "daily"
```
Abestanis, line 16:
```suggestion
      interval: "daily"

 - package-ecosystem: "github-actions"
    directory: "/"
    schedule:
      interval: "daily"

```
Let's also let it keep the actions in our CI up to date.

Abestanis:
Don't we want to enable the the `Dependabot version upgrades`? The point was to get some automatic PRs when there are new versions.

nt4f04uNd:
> Don't we want to enable the the `Dependabot version upgrades`? The point was to get some automatic PRs when there are new versions.

This PR is what it takes to enable it :-)

Abestanis:
> This PR is what it takes to enable it :-)

Ah, I didn't realize that. Perfect.

.github/dependabot.yml
Abestanis: [Docs](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot#example-dependabotyml-file-for-github-actions)

Abestanis (approved):

PR #181, opened 2024-10-27
Remove test flakiness

### LINKED ISSUE #162: Investigate test flakiness
https://github.com/nt4f04uNd/sweyer/actions/runs/11412091655/job/31757501525?pr=161

Possible solution
https://github.com/nt4f04uNd/sweyer/pull/161#issuecomment-2423351681

```
══╡ EXCEPTION CAUGHT BY FLUTTER TEST FRAMEWORK ╞════════════════════════════════════════════════════
The following LateError was thrown running a test (but after the test had completed):
LateInitializationError: Field '_onQueueChangeSubject@38260717' has not been initialized.

When the exception was thrown, this was the stack:
#0      QueueControl._onQueueChangeSubject (package:sweyer/logic/player/queue.dart)
#1      QueueControl.onQueueChanged (package:sweyer/logic/player/queue.dart:378:38)
#2      AudioHandler._init (package:sweyer/logic/player/music_player.dart:220:45)
#3      MusicPlayer.init (package:sweyer/logic/player/music_player.dart:647:14)
<asynchronous suspension>
#4      ContentControl.init (package:sweyer/logic/player/content.dart:248:9)
<asynchronous suspension>
#5      _SongsEmptyScreenState._handleRefetch (package:sweyer/routes/home_route/home_route.dart:[186](https://github.com/nt4f04uNd/sweyer/actions/runs/11412091655/job/31757501525?pr=161#step:3:198):5)
<asynchronous suspension>
════════════════════════════════════════════════════════════════════════════════════════════════════
❌ /home/runner/work/sweyer/sweyer/test/routes/home_route_test.dart: no songs screen - shows when the library is empty and pressing the button performs refetching (failed after test completion)
❌ /home/runner/work/sweyer/sweyer/test/routes/home_route_test.dart: no songs screen - shows when the library is empty and pressing the button performs refetching (failed after test completion)
══╡ EXCEPTION CAUGHT BY FLUTTER TEST FRAMEWORK ╞════════════════════════════════════════════════════
The following assertion was thrown running a test (but after the test had completed):
Current queue must not be empty
'package:sweyer/logic/player/queue.dart':
Failed assertion: line 257 pos 12: 'value.isNotEmpty'

When the exception was thrown, this was the stack:
#2      QueuesState.current (package:sweyer/logic/player/queue.dart:257:12)
#3      PlaybackControl.currentSongIndex (package:sweyer/logic/player/playback.dart:52:45)
#4      AudioHandler._setState (package:sweyer/logic/player/music_player.dart:616:44)
#5      AudioHandler._init.<anonymous closure> (package:sweyer/logic/player/music_player.dart:209:7)
#16     _StartWithStreamSink._safeAddFirstEvent (package:rxdart/src/transformers/start_with.dart:53:12)
#17     _StartWithStreamSink.onListen.<anonymous closure> (package:rxdart/src/transformers/start_with.dart:35:29)
(elided 20 frames from class _AssertionError, dart:async, and package:stack_trace)
════════════════════════════════════════════════════════════════════════════════════════════════════
❌ /home/runner/work/sweyer/sweyer/test/routes/home_route_test.dart: no songs screen - shows when the library is empty and pressing the button performs refetching (failed after test completion)
══╡ EXCEPTION CAUGHT BY FLUTTER TEST FRAMEWORK ╞════════════════════════════════════════════════════
The following assertion was thrown running a test (but after the test had completed):
Current queue must not be empty
'package:sweyer/logic/player/queue.dart':
Failed assertion: line 257 pos 12: 'value.isNotEmpty'

When the exception was thrown, this was the stack:
#2      QueuesState.current (package:sweyer/logic/player/queue.dart:257:12)
#3      PlaybackControl.currentSongIndex (package:sweyer/logic/player/playback.dart:52:45)
#4      AudioHandler._setState (package:sweyer/logic/player/music_player.dart:616:44)
#5      AudioHandler._init.<anonymous closure> (package:sweyer/logic/player/music_player.dart:215:44)
#16     Subject._add (package:rxdart/src/subjects/subject.dart:141:17)
#17     Subject.add (package:rxdart/src/subjects/subject.dart:135:5)
#39     _StartWithStreamSink._safeAddFirstEvent (package:rxdart/src/transformers/start_with.dart:53:12)
#40     _StartWithStreamSink.onListen.<anonymous closure> (package:rxdart/src/transformers/start_with.dart:35:29)
(elided 41 frames from class _AssertionError, dart:async, and package:stack_trace)
```

Abestanis:
This is an attempt to reduce the test flakiness. I noticed that `setUp` calls are not executed on the same `FakeAsync` zone than the tests, which means that they are more likely to leak into other tests and flushing mikro-tasks in the tests doesn't actually flush the ones scheduled in `setUp`. Maybe this is enough to fix the flakiness in the tests. It definitely helped to reveal two instances where we forgot to dispose a timer or animation controller.

I also removed all `runAsync` calls. As a general rule of thumb, we don't really want the tests to do any real async work, (network requests, file accesses or even worse, waiting for a timeout should all be mocked), so we should never use `runAsync`.

Fixes #162.


test/golden/goldens/artist_content_route.artist_content_route_songs.dark.png
Abestanis, line 1: Most of the changes in the golden are from the `AlbumArtRotating` widget which has a different rotation now that we don't have `runAsync` calls anymore before taking the screenshot. You can see the diff here, pixels that are different are painted red:

Golden diff with changes highlighted in red

test/golden/goldens/player_route.player_route.dark.artColor.png
Abestanis, line 1: This golden test has the only real difference: Because we no longer dispose the player before taking the screenshot, the seekbar is now full/at the end for the zero second song instead of empty/at the beginning.

Golden diff where changed pixels are red

Abestanis:
Given the fact that 100 cycles of the tests passed (see [CI run](https://github.com/nt4f04uNd/sweyer/actions/runs/11543757624/job/32128520987?pr=181) for 25134ad22afd6931bdc609885fe5636f1d7818ef) I'd say we can be pretty confident that this fixes the flakiness. 😁

test/routes/home_route_test.dart
```diff
@@ -146,27 +138,24 @@ void main() {
 
   testWidgets('no songs screen - shows when the library is empty and pressing the button performs refetching',
       (WidgetTester tester) async {
-    // TODO: because of `MusicPlayer.instance.stop` at the end of `runAppTest`, this test will print an error in console, but not actually fail, because the exception is caught. Ideally I should somehow hide that
```
Abestanis, line 149: There is no error printed anymore.

test/golden/goldens/artist_content_route.artist_content_route_songs.light.png
nt4f04uNd, line 1: Interestengly, a side effect I see on many screenshots is that the shadow from the drawer seems to have gone

It's OK, just something I noticed reviewing golden changes

test/test.dart
```diff
@@ -84,157 +86,192 @@ const kScreenWidth = 450.0;
...
+  ///  3. Pumps the app.
+  ///  4. Runs the test from the [callback].
+  ///  5. Optionally, runs [goldenCaptureCallback].
+  ///  6. Stops and disposes the player.
+  ///  7. Un-pumps the screen and flushes all micro-tasks and stream events.
+  Future<void> runAppTest(
+    AsyncCallback callback, {
+    AsyncCallback? goldenCaptureCallback,
+    VoidCallback? initialization,
+    VoidCallback? postInitialization,
```
nt4f04uNd, line 245: Should we align `runAppTestWithoutUi` with this one and pass this callback as well?

test/test.dart
```diff
@@ -84,157 +86,192 @@ const kScreenWidth = 450.0;
...
         await MusicPlayer.instance.dispose();
-      });
-      await goldenCaptureCallback?.call();
-      // Un-pump, in case we have any real animations running,
-      // so the pumpAndSettle on the next line doesn't hang on.
-      await pumpWidget(const SizedBox());
-      // Wait for ui animations.
-      await pumpAndSettle();
+        // Un-pump, in case we have any real animations running,
+        // so the pumpAndSettle on the next line doesn't hang on.
```
nt4f04uNd, line 268: This is no longer "next line"

test/test.dart
```diff
@@ -84,157 +86,192 @@ const kScreenWidth = 450.0;
...
+        await pumpWidget(const SizedBox());
+        // Wait for any asynchronous events and stream callbacks to finish.
+        await pump(const Duration(seconds: 1));
+        // Wait for ui animations.
```
nt4f04uNd, line 272: I imagine this is referring to "system UI" (but don't remember for sure)

Let's put that into the comment tho

Otherwise pumping out the UI above would already stop animations and we wouldn't need pumpAndSettle

test/test.dart
```diff
@@ -329,13 +369,23 @@ void testAppGoldens(
...
+          return tester.runAppTest(
+            initialization: () => initialization?.call(),
+            postInitialization: () {
+              ThemeControl.instance.setThemeLightMode(lightTheme);
+              Settings.playerInterfaceColorStyle.set(playerInterfaceColorStyle);
+              postInitialization?.call();
+            },
+            () => test(tester),
+            goldenCaptureCallback: () => tester.screenMatchesGolden(
+              Invoker.current!.liveTest.test.name.split(' | theme')[0].replaceAll(' ', '.'),
```
nt4f04uNd, line 385: Filename is already being built here https://github.com/nt4f04uNd/sweyer/blob/master/test/test.dart#L259

I'm don't complete understand the purpose of this code, but it seems like using the `description` parameter would already would already enrich it with the required parameters, maybe except from the group name

This code just seems to be not very scalable

test/logic/player/favorites_test.dart
```diff
@@ -10,46 +10,46 @@ void main() {
   final favoriteSong2 = songWith(id: 4, title: 'Song 4', isFavoriteInMediaStore: true);
   final favoriteSong3 = songWith(id: 5, title: 'Song 5', isFavoriteInMediaStore: true);
 
-  setUp(() async {
-    await setUpAppTest(() {
```
nt4f04uNd, line 14: Are we sure this is a good API change that we not longer can use setUp between tests?

test/routes/home_route_test.dart
nt4f04uNd: Nice

test/logic/player/favorites_test.dart
nt4f04uNd: I'd argue it's not - it reduces the reusability of setup logic

In some cases it might also make that tests utilities are not composable with each other - this is a real case in tests that I had recently, when I made a not very composable API desicion, which put constraints on how tests could be written, so I had to rewrite it

nt4f04uNd: From the description of the PR I see that you want them to live inside the same zone as other test logic

It this possible to achieve in some other way?

Abestanis:
Hey, I hope you're doing good and had a good start into the new year! 🎆

test/golden/goldens/artist_content_route.artist_content_route_songs.light.png
Abestanis: I don't actually see any difference in any shadow, the only diff I can see is in the `AlbumArtRotating` in the lower left because it has a slightly different angle. In which screenshot do you see that difference?
<details><summary>Difference</summary>

Pixel diff of this screenshot
</details>

test/test.dart
Abestanis: We could, but it would be really useless because we would call it directly before calling the callback. I don't think it's worth it:

```dart
Future<void> runAppTestWithoutUi(
    FutureOr<void> Function() callback, {
    VoidCallback? initialization,
    VoidCallback? postInitialization,
  }) async {
    addTearDown(postTest);
    return runTest(
      () => withClock(clock, () async {
        await _setUpAppTest(initialization);
        try {
          postInitialization?.call();
          await callback();
        } finally {
          // [...]
        }
      }),
      () {},
    );
  }
```

test/test.dart
```diff
@@ -84,157 +86,192 @@ const kScreenWidth = 450.0;
...
-  Future<void> runAppTest(AsyncCallback callback, {AsyncCallback? goldenCaptureCallback}) async {
+  /// Run a test that uses the App UI using the following procedure:
+  ///  1. Initializes the application, optionally running [initialization] when all fakes are set up
+  ///     before the [ContentControl] is initialized.
+  ///  2. Optionally, runs [postInitialization] after [ContentControl] is initialized.
+  ///  3. Pumps the app.
+  ///  4. Runs the test from the [callback].
+  ///  5. Optionally, runs [goldenCaptureCallback].
+  ///  6. Stops and disposes the player.
+  ///  7. Un-pumps the screen and flushes all micro-tasks and stream events.
```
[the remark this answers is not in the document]
Abestanis: Done.

test/test.dart
Abestanis: Yeah, I expanded the comment.

test/logic/player/favorites_test.dart
Abestanis: > It this possible to achieve in some other way?

Not that I can see. I also don't particularly like this, but the async zone is created by the test internals after the setUp is run, so there is no way for code that is run in `setUp` to be run in that zone. And the fact that we have futures running in a different zone is the main problem that causes the flakiness we are seeing.

We could try to have our own version of `setUp` that that somehow magically registers itself via some global and which we could run in `_setUpAppTest`, but this feels brittle and very complicated.

test/test.dart
Abestanis: This allows us to automatically deduce the golden filename from the test name, which I think is a big plus for us.

> ... maybe except from the group name

That's exactly the crux of the issue, we don't have the group information in the `description`, and the only way I could see to get it is via the test name.

> Filename is already being built here https://github.com/nt4f04uNd/sweyer/blob/master/test/test.dart#L259

It's not the same, that function expects as an argument the very string we are creating here.

> This code just seems to be not very scalable

Since we create the test description directly above, I changed it to use that description directly so it will still work if we change `TestDescription.buildDescription` in the future. The only other thing this relies on now is that the group comes before the test description in the test name, an I think it's worth accepting that assumption, especially since this is just used in the tests.

Abestanis:
The timeout on the CI is very weird, there were only changes in comments, and it runs fine [on my fork](https://github.com/Abestanis/sweyer/actions/runs/12736353906/job/35496150911?pr=2#step:3:221). Maybe something has changed on the CI. I'm looking into it.

nt4f04uNd:
> Hey, I hope you're doing good and had a good start into the new year! 🎆

Hey, yeah absolutely. I hope you had one too!

I've lost the context of most of the open PRs, so let me know if there are PRs that are waiting for my comments and which ones I should start with :-)

Abestanis:
> I've lost the context of most of the open PRs, so let me know if there are PRs that are waiting for my comments and which ones I should start with :-)

I need to check myself where we left of. 😄 I'll ping you in my other PRs when I refamiliarized myself with them, but I would really like this one to be merged because I hate the random test failures. I'll undraft once I removed the commit that tests 100 times.

This has been a very annoying issue in general, the changes on this branch worked well last year ([I tested by repeating the test step 100 times](https://github.com/nt4f04uNd/sweyer/commit/25134ad22afd6931bdc609885fe5636f1d7818ef)) and then in January it started failing because of timeouts without any code changes. So I spend the last few months trying to debug this every once in a while and I finally found the issue today.

This is only half of the actual fix, the real solution to this these problems we've been seeing is to not have singletons and use some actual dependency injection as you mentioned in #81. But this is still a step in the right direction.

test/logic/player/favorites_test.dart
Abestanis: I've added back support for `setUp`. It's using global state and depends on the fact that tests are not executed in parallel.

Abestanis:
Ok, this also passed 100 test (see [here](https://github.com/nt4f04uNd/sweyer/actions/runs/13875533245/job/38827336790)), so I am confident that in it's current state this does remove the flakiness. Let me know if you want any changes.

test/test.dart
Abestanis: I removed these arguments and switched to a registration based system via `registerAppSetup` and `registerPostAppSetup`.

nt4f04uNd:
@Abestanis can you add this context to #81 ?

Abestanis:
> @Abestanis can you add this context to #81 ?

Done

test/logic/player/music_player_test.dart
```diff
@@ -4,84 +4,93 @@ import 'package:rxdart/rxdart.dart';
...
   test('Player can pause', () async {
-    var playFutureCompleted = false;
-    MusicPlayer.instance.play().whenComplete(() => playFutureCompleted = true);
-    await pumpEventQueue();
-    expect(MusicPlayer.instance.playing, true);
-    await MusicPlayer.instance.pause();
-    await pumpEventQueue();
-    expect(playFutureCompleted, true);
-    expect(MusicPlayer.instance.playing, false);
+    final binding = TestWidgetsFlutterBinding.ensureInitialized();
```
nt4f04uNd, line 14: We shouldn't use WidgetsBinding inside a `test`, should we? If we use it's already a widget test and therefore should be `testWidgets`

test/test.dart
```diff
@@ -84,157 +86,216 @@ const kScreenWidth = 450.0;
...
-/// The [configureFakes] callback can be used to modify the fake data providers
-/// before the controls will load it.
-Future<void> setUpAppTest([VoidCallback? configureFakes]) async {
-  final binding = TestWidgetsFlutterBinding.ensureInitialized();
-  for (TestFlutterView view in binding.platformDispatcher.views) {
-    view.physicalSize = kScreenSize * kScreenPixelRatio;
-    view.devicePixelRatio = kScreenPixelRatio;
+extension AppInitExtension on TestWidgetsFlutterBinding {
+  /// Run a test that doesn't need the App UI using the following procedure:
+  ///  1. Initializes the application, optionally running callbacks registered with [registerException].
```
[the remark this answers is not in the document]
Abestanis: It's [a function from the `test_api` module](https://api.flutter.dev/flutter/flutter_test/registerException.html), I used autocomplete and didn't realized it completed to another function. Fixed.

test/test.dart
```diff
@@ -84,157 +86,216 @@ const kScreenWidth = 450.0;
...
+          child: const App(debugShowCheckedModeBanner: false),
+        ));
+        await pump();
+        await callback();
+        await goldenCaptureCallback?.call();
+      } finally {
+        // Don't leak app state between tests.
+        await MusicPlayer.instanceIfInitialized?.stop();
+        DeviceInfoControl.instance.dispose();
+        ContentControl.instance.dispose();
```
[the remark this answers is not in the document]
Abestanis: Good idea, done.

test/golden/routes_golden_test.dart
```diff
@@ -5,155 +5,131 @@ import '../observer/observer.dart';
...
-      }, goldenCaptureCallback: () => tester.screenMatchesGolden('tabs_route.selection_songs_tab'));
+      await tester.pumpAndSettle();
+      await tester.longPress(find.byType(SongTile));
+      await tester.pumpAndSettle();
     });
 
-    testAppGoldens('selection_deletion_dialog_songs_tab', (WidgetTester tester) async {
-      final List<Song> songs = List.unmodifiable(List.generate(10, (index) => songWith(id: index)));
-      await setUpAppTest(() {
+    final List<Song> songs = List.unmodifiable(List.generate(10, (index) => songWith(id: index)));
```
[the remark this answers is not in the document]
Abestanis: In general it's fine especially for non-mutable data, and this is the only way of sharing data in the setup callback and the main test callback. But in this instance it's not actually needed. I moved the initialization into the setup callback and removed the outside variable.

test/logic/player/music_player_test.dart
Abestanis: We've always done it (the call to `TestWidgetsFlutterBinding.ensureInitialized` was previously in `setUpAppTest`).
We need a binding to register our fakes and mock method call handlers, I'm not sure we can avoid it.

nt4f04uNd: Maybe we should just switch to `testWidgets` then?

Abestanis: Potentially, or investigate not needing a test binding. But I would prefer to do such a change in a separate PR to avoid feature creep here. I created #261 to track that.

nt4f04uNd (approved):

PR #262, opened 2025-03-17
Create SECURITY.md

nt4f04uNd:
A boilerplate file, just in case it ever be needed.
I decided to keep it short.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,5 +1,14 @@
 # Security Policy
 
+We take security seriously and strive to respond promptly and responsibly to all reported vulnerabilities.
+
 ## Reporting a Vulnerability
 
-To report a security issue, please use the GitHub Security Advisory ["Report a Vulnerability"](https://github.com/nt4f04uNd/sweyer/security/advisories/new) tab.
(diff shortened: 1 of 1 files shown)


SECURITY.md
```diff
@@ -0,0 +1,5 @@
+# Security Policy
+
+## Reporting a Vulnerability
+
+To report a security issue, please use the GitHub Security Advisory ["Report a Vulnerability"](https://github.com/nt4f04uNd/sweyer/security/advisories/new) tab.
```
Abestanis, line 5: I think this is only intended to be used by maintainers and other people can't create anything there. The link is definitely not accessible without logging in, so people without a GitHub account wouldn't be able to report anything.

Abestanis: Maybe just put an email address there instead?

Abestanis: Thanks for adding the email. I checked and according to [the docs](https://docs.github.com/en/code-security/security-advisories/guidance-on-reporting-and-writing-information-about-vulnerabilities/privately-reporting-a-security-vulnerability) it is possible for non-admins to report vulnerabilities, so this is perfect.

Abestanis (approved):

PR #274, opened 2025-03-30
Update to Flutter 3.29.2

Abestanis:
Update to the newest Flutter version. This also updates all dependencies, but does not update any major versions, to keep this MR short (except `audio_service`). This also does not fix the new warnings. During build, there is a new warning:
```
Synthetic package output (package:flutter_gen) is deprecated: https://flutter.dev/to/flutter-gen-deprecation. In a future release, synthetic-package will default to `false`
and will later be removed entirely.
```
I'll address this in a future PR.

This PR depends on https://github.com/nt4f04uNd/nt4f04unds_widgets/pull/14, but is already ready for review.

### DIFF
--- a/lib/logic/player/music_player.dart
+++ b/lib/logic/player/music_player.dart
@@ -181,6 +181,11 @@ class _BrowserParentProvider {
 
 @visibleForTesting
 class AudioHandler extends BaseAudioHandler with SeekHandler, WidgetsBindingObserver {
+  @visibleForTesting
+  static const loopOn = 'loop_on';
+  @visibleForTesting
+  static const loopOff = 'loop_off';
+
   AudioHandler(MusicPlayer player) {
     _init(player);
   }
@@ -433,7 +438,7 @@ class AudioHandler extends BaseAudioHandler with SeekHandler, WidgetsBindingObse
   }
 
   @override
-  Future<void> setRating(Rating rating, Map? extras) {
+  Future<void> setRating(Rating rating, [Map<String, dynamic>? extras]) async {
     // TODO: implement setRating
     throw UnimplementedError();
   }
@@ -497,20 +502,11 @@ class AudioHandler extends BaseAudioHandler with SeekHandler, WidgetsBindingObse
   }
 
   @override
-  Future<void> onNotificationAction(String action) async {
-    switch (action) {
-      case 'loop_on':
-      case 'loop_off':
+  Future<dynamic> customAction(String name, [Map<String, dynamic>? extras]) {
+    switch (name) {
+      case loopOn:
+      case loopOff:
         return player.switchLooping();
-      case 'play_prev':
-        return player.playPrev();
-      case 'pause':
-      case 'play':
-        return player.playPause();
-      case 'play_next':
-        return player.playNext();
-      case 'stop':
-        return stop();
       default:
         throw UnimplementedError();
     }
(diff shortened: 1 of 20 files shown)


lib/routes/home_route/player_route.dart
```diff
@@ -878,7 +878,7 @@ class _InfoButton extends StatelessWidget {
             ),
           ),
           additionalActions: [
-            CopyButton(text: songInfo),
+            IntrinsicWidth(child: CopyButton(text: songInfo)),
```
Abestanis, line 881: See https://github.com/nt4f04uNd/nt4f04unds_widgets/pull/14#discussion_r2020273229

lib/widgets/show_functions.dart
```diff
@@ -171,7 +171,7 @@ class ShowFunctions extends NFShowFunctions {
                   ),
                 ),
                 additionalActions: [
-                  CopyButton(text: errorDetails),
```
Abestanis, line 174: See https://github.com/nt4f04uNd/nt4f04unds_widgets/pull/14#discussion_r2020273229

test/logic/player/music_player_test.dart
```diff
@@ -75,19 +84,23 @@ void main() {
       await binding.runAppTestWithoutUi(() async {
         MusicPlayer.instance.play();
         await binding.pump();
-        await MusicPlayer.handler!.onNotificationAction('stop');
+        await MusicPlayer.handler!.stop();
```
Abestanis, line 87: Clicking the `stop` action on the notification ends up calling `stop`, so this is equivalent to the behaviour from before.

test/logic/player/music_player_test.dart
```diff
@@ -75,19 +84,23 @@ void main() {
...
         await binding.pump();
         final playbackState = MusicPlayer.handler!.playbackState.value!;
         expect(playbackState.playing, false);
-        expect(playbackState.processingState, AudioProcessingState.ready);
```
Abestanis, line 82: This has changed in `audio_service`, I verified that calling `stop` should cause an idle processing state.

pubspec.yaml
```diff
@@ -78,8 +77,8 @@ dependencies:
     # path: /Users/nt4f04und/dev/prj/nt4f04unds_widgets
 
     git:
-      url: https://github.com/nt4f04uNd/nt4f04unds_widgets
-      ref: 4.0.0
+      url: https://github.com/Abestanis/nt4f04unds_widgets
+      ref: update/flutter
```
Abestanis, line 81: TODO: This needs to be updated before merging.

lib/logic/player/music_player.dart
```diff
@@ -531,45 +529,44 @@ class AudioHandler extends BaseAudioHandler with SeekHandler, WidgetsBindingObse
     final l10n = staticl10n;
     playbackState.add(playbackState.value!.copyWith(
       controls: [
-        // TODO: currently using custom API from my fork, see https://github.com/ryanheise/audio_service/issues/633
```
nt4f04uNd, line 534: The issue remains open. Did they actually support what this Todo was referring to?

I've lost almost all context hete

lib/logic/player/music_player.dart
```diff
@@ -601,14 +598,14 @@ class AudioHandler extends BaseAudioHandler with SeekHandler, WidgetsBindingObse
       androidCompactActionIndices: const [1, 2, 3],
       repeatMode: player.looping ? AudioServiceRepeatMode.one : AudioServiceRepeatMode.all,
       shuffleMode: QueueControl.instance.state.shuffled ? AudioServiceShuffleMode.all : AudioServiceShuffleMode.none,
-      processingState: const {
-        ProcessingState.idle: AudioProcessingState.idle,
-        ProcessingState.loading: AudioProcessingState.loading,
-        ProcessingState.buffering: AudioProcessingState.buffering,
-        ProcessingState.ready: AudioProcessingState.ready,
-        ProcessingState.completed: AudioProcessingState.completed,
-        // Excluding idle state because it makes notification to reappear.
```
nt4f04uNd, line 610: So was this fixed or why can we rewrite this now?

lib/logic/player/music_player.dart
```diff
@@ -667,7 +664,6 @@ class MusicPlayer extends AudioPlayer {
         // artDownscaleHeight,
         fastForwardInterval: const Duration(seconds: 5),
         rewindInterval: const Duration(seconds: 5),
-        androidEnableQueue: true,
```
nt4f04uNd, line 670: Is this now enabled by default?

lib/logic/player/music_player.dart
Abestanis: As far as I can tell, yes. We are using it to add an additional action (loop on/off) that is not built-in to audio_service, which is what the issue was about if I'm not mistaken.

lib/logic/player/music_player.dart
Abestanis: Yes. see [this warning in their code](https://github.com/ryanheise/audio_service/blob/60c6dfd24d2b571fe0a0c77fbb08daeba259b33e/audio_service/lib/audio_service.dart#L1359):
```
print('NOTE: androidEnableQueue is always true from 0.18.0 onwards.');
```

lib/logic/player/music_player.dart
Abestanis: No, the code is identical, I just switched it from using a map to using a `switch`, but I forgot to add the comment back.

Abestanis: I cleaned it up and added back the comment, but I adjusted it a bit. Please let me know if it's ok like that.

lib/logic/player/music_player.dart
nt4f04uNd: Perfect, closed the issue then

nt4f04uNd (approved):

PR #294, opened 2025-04-06
Add formatting and analyse CI checks

### LINKED ISSUE #273: Add analyze + format checks on CI
### Suggestion

Same as https://github.com/nt4f04uNd/sweyer/issues/272, but for CI

### LINKED ISSUE #275: Rename github action steps
### Suggestion

This is confusing

Image

Abestanis:
Resolves #273 and fixes #275.

This is similar to https://github.com/nt4f04uNd/nt4f04unds_widgets/pull/15. I checked that both checks pick up problems in the `sweyer_plugin` project as well.

Example CI runs:
* [With formatting errors](https://github.com/nt4f04uNd/sweyer/actions/runs/14297702087)
* [With analyze errors](https://github.com/nt4f04uNd/sweyer/actions/runs/14297735340)

### DIFF
--- a/.github/problem-matchers/flutter-analyze.json
+++ b/.github/problem-matchers/flutter-analyze.json
@@ -1,0 +1,32 @@
+{
+  "problemMatcher": [
+    {
+      "owner": "flutter-analyze",
+      "pattern": [
+        {
+          "regexp": "^\\s*(error|warning) • (.*?) • (.*):([0-9]+):([0-9]+) • (.*)$",
+          "severity": 1,
+          "message": 2,
+          "file": 3,
+          "line": 4,
+          "column": 5,
+          "code": 6
+        }
+      ]
+    },
+    {
+      "owner": "flutter-analyze-info",
+      "severity": "notice",
+      "pattern": [
+        {
+          "regexp": "^\\s*info • (.*?) • (.*):([0-9]+):([0-9]+) • (.*)$",
+          "message": 1,
+          "file": 2,
+          "line": 3,
+          "column": 4,
+          "code": 5
+        }
+      ]
+    }
+  ]
+}
--- a/.github/workflows/build_executables.yml
+++ b/.github/workflows/build_executables.yml
@@ -23,9 +23,7 @@ jobs:
 
       - name: 📦 Install Dependencies
         shell: bash
-        run: |
-          flutter pub global activate very_good_cli
-          very_good packages get --recursive
+        run: flutter pub get
 
       - name: 💾 Cache Gradle
         uses: actions/cache@v4
--- a/.github/workflows/build_executables.yml
+++ b/.github/workflows/build_executables.yml
@@ -8,7 +8,7 @@ on:
       - 'releases/**'
 
 jobs:
-  build:
+  build_executables:
     runs-on: ubuntu-latest
     steps:
       - name: 📚 Git Checkout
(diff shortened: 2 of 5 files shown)


.github/workflows/test/action.yml
```diff
@@ -31,8 +28,23 @@ runs:
...
+      run: |
+        echo "::add-matcher::${{ github.workspace }}/.github/problem-matchers/dart-format.json"
+        dart format --set-exit-if-changed .
+        echo "::remove-matcher owner=dart-format::"
+
+    - name: 📝 Run Dart Linter
+      shell: bash
+      run: |
+        echo "::add-matcher::${{ github.workspace }}/.github/problem-matchers/flutter-analyze.json"
+        flutter analyze --fatal-infos || echo "Ignoring analyze problems for now" # TODO: Remove the `|| ...`
```
Abestanis, line 45: Fixing the remaining warnings and notes is going to take some time, I think it's better if we have the annotations anyways and make them fatal later.

.github/workflows/test/action.yml
```diff
@@ -31,8 +28,23 @@ runs:
     - name: 📦 Install Dependencies
       shell: bash
       run: |
-        flutter pub global activate very_good_cli
-        very_good packages get --recursive
+        flutter pub get
+        flutter pub get -C sweyer_plugin
```
nt4f04uNd, line 32: If we add a new package, it will not be picked up here

.github/problem-matchers/flutter-analyze.json
```diff
@@ -0,0 +1,32 @@
...
+          "code": 6
+        }
+      ]
+    },
+    {
+      "owner": "flutter-analyze-info",
+      "severity": "notice",
+      "pattern": [
+        {
+          "regexp": "^\\s*info • (.*?) • (.*):([0-9]+):([0-9]+) • (.*)$",
```
nt4f04uNd, line 22: Maybe we should make it failing, and instead use ignores inside the code to suppress, if we feel we need to fix it later

.github/workflows/build_executables.yml
```diff
@@ -23,9 +23,7 @@ jobs:
 
       - name: 📦 Install Dependencies
         shell: bash
-        run: |
-          flutter pub global activate very_good_cli
-          very_good packages get --recursive
```
nt4f04uNd, line 28: Why no longer use it?

.github/workflows/test/action.yml
Abestanis: `flutter pub get` does run recursively in sub-directories.
The only reason this is necessary is because it doesn't run in the `example` folder in the subdirectories, which causes a lint error. So if we do ever add add another package we'll immediately get a lint error to remind us, which is why I think it's acceptable.

.github/problem-matchers/flutter-analyze.json
Abestanis: It's much easier to forget suppressed warnings like that. We do want to fix them all, they are 2 actual valid warnings plus a bunch of deprecation notices, which will eventually turn into errors.

If you are concerned about the annotations added to all future PRs, then maybe we should wait until all of these warnings have been fixed before merging this.

.github/workflows/build_executables.yml
Abestanis: It's totally unnecessary and just bloats the CI, especially in this workflow a simple `flutter pub get` does exactly what we need.

nt4f04uNd (approved):

Abestanis:
@nt4f04uNd I think we need to temporarily disable he branch protection rule that requires the old workflows to succeed before merging and then re-enable them again with the new names. I don't have access to the relevant settings, could you please change them and merge this PR?

PR #299, opened 2025-04-08
This fixes the warnings from `flutter analyze` that are easy to fix

Abestanis:
This is a preparation to have the CI fail if `flutter analyze` reports any issues (#294). The other issues are harder to fix, I'll try to tackle them in other PRs.

### DIFF
--- a/lib/constants/theme/app_theme.dart
+++ b/lib/constants/theme/app_theme.dart
@@ -65,8 +65,8 @@ class AppTheme extends ThemeExtension<AppTheme> with _$AppThemeTailorMixin {
     currentIndicatorBackgroundColorWithDefaultArt: constants.Theme.defaultPrimaryColor,
     drawerMenuItemColor: const Color(0xff3d3e42),
     glowSplashColor: constants.Theme.lightThemeSplashColor,
-    glowSplashColorOnContrast: Colors.white.withOpacity(0.13),
-    sliderInactiveColor: Colors.black.withOpacity(0.2),
+    glowSplashColorOnContrast: Colors.white.withValues(alpha: 0.13),
+    sliderInactiveColor: Colors.black.withValues(alpha: 0.2),
   );
 
   static final AppTheme dark = AppTheme(
@@ -76,8 +76,8 @@ class AppTheme extends ThemeExtension<AppTheme> with _$AppThemeTailorMixin {
     contrastInverse: Colors.black,
     currentIndicatorBackgroundColorWithDefaultArt: constants.Theme.defaultPrimaryColor,
     drawerMenuItemColor: Colors.white,
-    glowSplashColor: Colors.white.withOpacity(0.1),
-    glowSplashColorOnContrast: Colors.black.withOpacity(0.13),
-    sliderInactiveColor: Colors.white.withOpacity(0.2),
+    glowSplashColor: Colors.white.withValues(alpha: 0.1),
+    glowSplashColorOnContrast: Colors.black.withValues(alpha: 0.13),
+    sliderInactiveColor: Colors.white.withValues(alpha: 0.2),
   );
 }
--- a/lib/utils/color.dart
+++ b/lib/utils/color.dart
@@ -8,14 +8,9 @@ import 'package:flutter/material.dart';
 /// Inspired by https://stackoverflow.com/a/13532993
 Color shadeColor(double factor, Color color, {int minBit = 0, int maxBit = 0xff}) {
   assert(factor >= -1 && factor <= 1);
-  var r = ((color.value >> 16) & 0xff).clamp(0, 0xff);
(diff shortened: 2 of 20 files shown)


lib/constants/theme/app_theme.dart
```diff
@@ -65,8 +65,8 @@ class AppTheme extends ThemeExtension<AppTheme> with _$AppThemeTailorMixin {
     currentIndicatorBackgroundColorWithDefaultArt: constants.Theme.defaultPrimaryColor,
     drawerMenuItemColor: const Color(0xff3d3e42),
     glowSplashColor: constants.Theme.lightThemeSplashColor,
-    glowSplashColorOnContrast: Colors.white.withOpacity(0.13),
-    sliderInactiveColor: Colors.black.withOpacity(0.2),
+    glowSplashColorOnContrast: Colors.white.withValues(alpha: 0.13),
+    sliderInactiveColor: Colors.black.withValues(alpha: 0.2),
```
nt4f04uNd, line 69: It's not an equal conversion

nt4f04uNd (changes requested):

lib/utils/color.dart
```diff
@@ -8,14 +8,9 @@ import 'package:flutter/material.dart';
...
-
   factor = (1 + factor);
-  r = (r * factor).toInt().clamp(minBit, maxBit);
-  g = (g * factor).toInt().clamp(minBit, maxBit);
-  b = (b * factor).toInt().clamp(minBit, maxBit);
-
-  return Color((0xff << 24) + (r << 16) + (g << 8) + b);
+  final r = (color.r * 255.0 * factor).toInt().clamp(minBit, maxBit);
+  final g = (color.g * 255.0 * factor).toInt().clamp(minBit, maxBit);
+  final b = (color.b * 255.0 * factor).toInt().clamp(minBit, maxBit);
```
nt4f04uNd, line 14: Seems like we no logner need int conversion here

lib/utils/color.dart
```diff
@@ -8,14 +8,9 @@ import 'package:flutter/material.dart';
...
+  return Color.fromARGB(0xFF, r, g, b);
```
nt4f04uNd, line 15: Can you make an issue to check whether we should put an original alpha color here instead? I'm not sure I did this intentionally or not when I implemented this

lib/constants/theme/app_theme.dart
Abestanis: Are you sure? What makes you say that? From what I can see:
* `withOpacity(x)` -> `withAlpha((255.0 * x).round())` -> `Color.fromARGB((255.0 * x).round(), ...)` -> `this._fromRGBOC(..., ((255.0 * x).round() & 0xff) / 255, colorSpace)` -> `a = ((255.0 * x).round() & 0xff) / 255` which is basically `a = x`
* `withValues(alpha: x)` -> `a = x`

nt4f04uNd: Yep, you are right

lib/utils/color.dart
Abestanis: `fromARGB` takes an `int`, and without the `toInt` these expressions produce a `num`.

But maybe it would be better if we convert the `minBit` and `maxBit` into `0.0`-`1.0` domain and just use floats everywhere. The only reason why I didn't do it is because we loose the `floor` operation performed by `toInt`, so the results will be slightly different from before.

lib/utils/color.dart
Abestanis: In my opinion it should not change the opacity. Let's use the opacity form he color argument, and let's update the golden test images and see if we have any changes.

lib/utils/color.dart
nt4f04uNd: Are you sure we still need to use `fromARGB` and not `from`?

lib/utils/color.dart
nt4f04uNd: Yes, but I would prefer if we did that in a separate issue

lib/utils/color.dart
Abestanis: `from` takes values in the `0.0` to `1.0` domain. We can use that if we avoid the multiplication by `255` and if we convert the `minBit` and `maxBit` to floats in that domain. I think it's the better solution, but it will change the output of this function slightly, since we won't have the `floor` from the `toInt` anymore.

Should we also move this into a separate issue?

Abestanis: Or maybe the same issue about the alpha channel?

nt4f04uNd: Let's do these 2 refactors in the separate issue and leave it as-is in this PR

nt4f04uNd (approved):

Issue #267, opened 2025-03-25
Flutter 3.29.0


nt4f04uNd:
### Suggestion

We are currently blocked by https://github.com/flutter/flutter/issues/163627

nt4f04uNd:
After update update these dependencies:
https://github.com/nt4f04uNd/sweyer/pull/247
https://github.com/nt4f04uNd/sweyer/pull/170

Abestanis:
I've already started working on it and I did find a good workaround for https://github.com/flutter/flutter/issues/163627, but I won't have time before this weekend to finish it.
See https://github.com/Abestanis/sweyer/tree/update/flutter and https://github.com/Abestanis/nt4f04unds_widgets/tree/update/flutter.

nt4f04uNd:
Good! There's no rush, of course this could wait even much longer than just a week

Abestanis:
Done in #274.

PR #310, opened 2025-04-10
Update Golden test artifacts

Abestanis:
I forgot to update those in #267. I didn't see any significant changes in the screenshots.

Also update `sweyer_plugin/example/pubspec.lock`, which now get's updated on CI


nt4f04uNd:
How come this didn't break the master?

Abestanis:
> How come this didn't break the master?

Because the changes are all minimal, a couple pixels at max and in some of them there is no pixel change at all and I can't really tell why the file is different.

And since we allow a small percentage to be different, this didn't cause the golden tests to fail.

Maybe we should stop allowing this small difference? It doesn't really help me with running them locally, on my machine the outputs are different enough to cause some of the Goldens to fail.
If you think it's valuable to keep allowing that difference, maybe we can just disallow it on the CI?

nt4f04uNd:
> And since we allow a small percentage to be different, this didn't cause the golden tests to fail

Do we? Well, it doesn't seem good, I don't remember how and when we enabled this, but this can lead to inconsistency in the long run, so I would reconsider to disallow it

nt4f04uNd (approved):

Abestanis:
Created #312 to track removing the tolerance.

PR #316, opened 2025-04-12
Don't use a synthetic generated package for localizations

Abestanis:
This fixes the following warning on every build:
```
Synthetic package output (package:flutter_gen) is deprecated: https://flutter.dev/to/flutter-gen-deprecation. In a future release, synthetic-package will default to `false` and will later be removed entirely.
```

I've followed the [migration guide](https://docs.flutter.dev/release/breaking-changes/flutter-generate-i10n-source#migration-guide), now the sources are generated into `lib/localization/generated` instead of a synthetic package.

### DIFF
--- a/.github/workflows/test/action.yml
+++ b/.github/workflows/test/action.yml
@@ -31,6 +31,10 @@ runs:
         flutter pub get
         flutter pub get -C sweyer_plugin
 
+    - name: 🌐 Generate localizations
+      shell: bash
+      run: flutter gen-l10n
+
     - name: 📄 Check Dart formatting
       shell: bash
       run: |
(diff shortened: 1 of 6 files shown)


nt4f04uNd (approved):

nt4f04uNd:
What happens if one tries to launch the app build without generating localizations?

Abestanis:
It's generated automatically when calling `flutter build` (like the `build_executable` workflow). Unfortunately they are not generated by `flutter analyze` and `flutter test`.

Abestanis:
Ah, but you are actually onto something: These should be generated automatically when running `flutter pub get` according to https://github.com/flutter/flutter/pull/132172, but they are not. Apparently we still want the `generate: true` line in the pubspec. I'll create a PR for that.

PR #317, opened 2025-04-13
Let flutter automatically generate localization files on `pub get`

Abestanis:
See https://github.com/nt4f04uNd/sweyer/pull/316#issuecomment-2799912277. Without the `generate: true`, flutter only generates them on `flutter build`. I thought in #316 that we need to get rid of this  setting to fix the warning, but it looks like we want it and it's ok to have it as long as we are not using the synthetic package.

### DIFF
--- a/.github/workflows/test/action.yml
+++ b/.github/workflows/test/action.yml
@@ -31,6 +31,10 @@ runs:
         flutter pub get
         flutter pub get -C sweyer_plugin
 
+    - name: 🌐 Format generated localizations
+      shell: bash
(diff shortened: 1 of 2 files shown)


Abestanis:
Great, it looks like the format option is not respected when generating via `flutter pub get`. Looks like there is some work done on flutter right now which might fix this (they are enabling formatting by default) [here](https://github.com/flutter/flutter/pull/167029), so I'll add a step to format them manually for now, hopefully we can remove it on the next Flutter update.

nt4f04uNd (approved):

PR #318, opened 2025-04-13
Update the project template to adjust to the latest Flutter version

Abestanis:
This also increases the `compileSdk` version from `34` to `35`, but the `targetSdk` remains the same. We can update it later.


Abestanis:
This bumps the minimum dart version, which enables a formatter change ([described here](https://github.com/dart-lang/dart_style/issues/1253)): The formatter now automatically removes/adds commas in argument lists. This removes our ability to manually influence the formatting there (There will be [an option to bring back that control](https://github.com/dart-lang/dart_style/pull/1672), but it is has only been merged last month, so well need to wait at least until the next Flutter update to be able to use it).

I think most of these formatting changes are good, although some of them seem a bit worse. But this removes one more thing that we need to be aware when writing and reviewing code, and it will enforce a uniform and consistent formatting, so I'm in favour of this change. What do you think?

pubspec.yaml
```diff
@@ -7,8 +7,7 @@ publish_to: 'none'
 version: 1.0.13+16
 
 environment:
-  sdk: '>=3.4.1 <4.0.0'
-  flutter: 3.29.2
```
nt4f04uNd, line 11: Why remove flutter entry? I specifically put it into here for anyone running the project to be aware of the targeting Flutter version

nt4f04uNd:
Let's wait for the fix from the dart team and bump it then. There are too many questionable style-preference changes here.

pubspec.yaml
Abestanis: Because the default template does not have it, I guess it is not common for apps to fix the Flutter version like that. But I'll  can add it back.

Abestanis:
> Let's wait for the fix from the dart team and bump it then. There are too many questionable style-preference changes here.

Ok, then I'll drop the dart sdk version and undo the formatting changes.

pubspec.yaml
Abestanis: Added it back

nt4f04uNd (approved):

PR #380, opened 2025-10-05
Disable printing of errors reported to Firebase in the tests

Abestanis:
Reporting an error to Firebase prints it by default (`printDetails` is `true` if not specified). These prints can make it really hard to read anything in the test logs, because we have some tests that trigger hundreds of them. Let's disable the print in the tests.


nt4f04uNd (approved):

lib/utils/firebase.dart
```diff
@@ -0,0 +1,33 @@
+import 'package:firebase_crashlytics/firebase_crashlytics.dart';
+import 'package:flutter/material.dart';
+
+@visibleForTesting
+bool printFirebaseErrors = true;
```
nt4f04uNd, line 5: I guess the other way of doing it would be to mock FirebasePluginPlatform instance and rewrite CrashlyticsObserver to use it, but this also works

nt4f04uNd:
Oh, I though this PR wasn't reviewed, but now I see it is approved! Let's merge it

Abestanis:
> Let's merge it

Ok, I've just merged main into it to fix the CI error.

